PR #10, opened 2016-12-21
Bridges

flowlo:
Hi!

This is not a semantic change as to how querying of external atoms is implemented, but an important architectural one.

*Implementing a new Grounder and Solver pair that violate the existing (generic) interface will produce **major pain** down the road.*

While I understand that you paved the way for further extending `NaiveGrounder`, I think that composition would have more benefits. Querying an external source of atoms is just one added possibility for any grounder to obtain new nogoods. Also, it should be possible to query hex from any grounder implementation that might pop up. Further inheritance would be horrible, while a composable approach could just be reused.

Therefore I propose to add a separate entity which can be composed into any grounder. Call that new entity a *Bridge* (because it bridges between the grounder and some other source of nogoods, for the lack of a better name; suggestions welcome).

It models *any external source of nogoods* that can be queried, given the current atoms and their assignment.

As bridges are composable, any grounder can query any number of bridges generically.

What this change does:
 1. Add `Bridge`. It models what a bridge is.
 2. The functionality contained `AlphahexGrounder` is moved to `HexBridge implements Bridge`.
 3. A new subtype `BridgedGrounder extends AbstractGrounder` adds boilerplate for querying any number of bridges supplied at construction.

What will break:
 1. As I moved `native` methods to an appropriate namespace (I *strongly object* to further cluttering `Main`), you will have to rename exports on hex' side accordingly. This should be straightforward, [see the JNI Specification](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/design.html#resolving_native_method_names).
 2. The direct call to `sendResults` in`Main` is not possible anymore because of (1.).
 3. Testing just got much harder.

What remains unclear/sketchy:
 1. When exactly to call `collectExternal()` in `NativeGrounder`. I only have a rough idea.
 2. How to obtain the current assignment.
 3. How to construct bridges. Probably we'll have (yet another) command line flag to enable them.
 4. When/how to call `sendResults`?

If you (@tobiaskaminski, @AntoniusW) are content with the proposed changes, I am happy to help put together the missing pieces as well.


flowlo:
> We could call sendResults of HexBridge in the end of main whenever this option is set. I think sendResults should be called in main, so that Hex calls the main method (being able to provide different options) and receives all answer sets computed by Alpha directly before Alpha finishes.

I think I have yet to understand what `sendResults` actually does, and how Alpha/Hex relate to each other. Am I right that Hex spawns Alpha, feeds it with nogoods until all ASs have been found and then consumes the ASs? If that's the case there is no nice symmetry with the bridge, but we could rather define a second "output format" of answer sets, again triggered by a command line switch (which might be the same that activates the Hex bridge). I think I'll try it that way. Alpha will get a "run by Hex" flag that will set up the Bridge and report back answer sets.

> I don't see any reason why not simply to call the getNoGoods methods of the Bridges at the end of getNoGoods of the respective grounder.

Great, that's where I thought the call should be. Right where I placed the `TODO`?

> One idea of deriving a new grounder and solver class also was that the assignment itself is not needed in NaiveGrounder. But if we consider arbitrary nogood providers in the grounder, we need access to the assignment anyway.

Yes, this is is something @AntoniusW and me (very briefly) discussed. I have yet to find a clean solution... :smile:

I'll dab around some more and go ahead with merging.

flowlo:
I added a flag (`OPT_HEX`) in `Main` that will trigger initialisation of a `HexBridge` and also lead to not reporting answer sets via standard output but call `HexBridge.sendResults`.

As for passing the current `Assignment` over to the grounder, I added `ImmutableAssignment` which is a supertype of `Assignment`. It contains all "read-only" methods of what we had in `Assignment` up to now. By using this more restrictive supertype when passing over the assignment to the grounder, we can make sure that the grounder only reads, but does not modify the assignment (which, to my understanding is something only the solver is allowed to do).

To remove the cast to `BasicAssignment` I introduced the interface `AssignmentFormatter` which is there to model arbitrary formatting transformations on an answer set. We have the default one, that just calls the respective `toString` Method, and one for Hex, that respects the format that Hex is used to. Again, they are composable.

I do not have Hex, so I did not test these changes end-to-end, though I suspect everything to work out as soon as exported symbol names are change in Hex. Making it work probably is more relevant when merging with `master`.

tobiaskaminski:
Thanks a lot for helping with the architecture! I have made the changes in Hex, and everything works fine :).

Regarding your question about the Alpha/Hex relation:

The Hex algorithm roughly works by splitting a Hex program (ASP+external atoms) into different units, transforming them into ordinary ASP programs and using an ordinary ASP solver to retrieve their answer sets. The general idea of Alphahex is to use Alpha as a _backend_ solver, where the evaluation of external atoms is interleaved with solving in Alpha.

The crucial point is that the truth value of external atoms are not defined in terms of rules, but depend globally on the assignment to the ordinary atoms in the program. Usually, external atoms only depend on a part of the assignment constituted by the extension of those predicates that are declared as their input. However, in general the concrete dependencies are hidden from the ASP solver (e.g. external atoms may behave nonmonotonically). For this reason, after propagating, Alpha provides its current partial assignment to Hex, which checks for new external atom assignments that are implied by the partial assignment and enforces the corresponding assignments in Alpha by adding nogoods. After Alpha has returned all answer sets of a unit, they are integrated over the different components of the Hex program, and its answer sets are returned by Hex.

PR #13, opened 2016-12-23
Alphahex

lorenzleutgeb:


flowlo:
This has to be delayed until after #15

flowlo:
@tobiaskaminski I performed a control merge from `master` and restructured the way choices are stored a bit (this will come in handy on `master` as well). I'd like to ask you (again) to verify the setup in connection with Hex. I hope I did not break anything along the way ...

If your tests turn out well, I think we should merge this ASAP.

There's more interesting stuff to come in [`bridges`](https://github.com/AntoniusW/Alpha/compare/bridges).

flowlo:
OK. Nice to hear that comms with Hex are looking good!

AntoniusW:
The exception should not occur, but I cannot determine the source of the likely bug from just looking at the Hex program. The exception indicates that during learning, it resolved until the first UIP (which usually is assigning) but the NoGood was not assigning, hence the backjumping level could not be computed and it returned -1 instead. Is it possible to simulate the run just inside Alpha (i.e., which NoGoods were added by Hex)?

flowlo:
I'll try to come up with a reproduction for that.

flowlo:
I did a control merge and am now getting following trace for many tests.

```
java.lang.NullPointerException
	at at.ac.tuwien.kr.alpha.solver.ChoiceManager.backtrack(ChoiceManager.java:96)
	at at.ac.tuwien.kr.alpha.solver.DefaultSolver.doBacktrack(DefaultSolver.java:256)
	at at.ac.tuwien.kr.alpha.solver.DefaultSolver.tryAdvance(DefaultSolver.java:129)
	at at.ac.tuwien.kr.alpha.solver.AbstractSolver$1.tryAdvance(AbstractSolver.java:31)
	at java.util.Spliterator.forEachRemaining(Spliterator.java:326)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
	at at.ac.tuwien.kr.alpha.solver.Solver.collectSet(Solver.java:21)
```

flowlo:
@tobiaskaminski In order to get this branch stable, I resorted to reverting your commit 9ca29c0 (see the difference of test runs between 068f918 and 3944f8a). Now everything should work as expected, but the branch is missing this specific changeset. Let's merge this now and get your changes in separately, or do you prefer sorting that out before merging?

tobiaskaminski:
@AntoniusW , for the example program, Hex does not return any nogoods before `assignUnassignedToFalse()` is called, and after it has been called the first time always returns the nogoods {{"aux_n_10000000_2a(a,c)", "-a(c)"}, {"aux_n_10000000_2a(b,c)", "-b(c)"}}. The identical behavior can be simulated without Hex by replacing `String[][] externalNogoods = getExternalNoGoods(assignment, atomStore);` in line 18 of `HexBridge.java` with the following, in order to add the same nogoods as Hex would return:
```
String[][] externalNogoods1 = {{"aux_n_10000000_2a(a,c)", "-a(c)"}, {"aux_n_10000000_2a(b,c)", "-b(c)"}};
String[][] externalNogoods2 = {};
String[][] externalNogoods;
if(assignment.toString().equals("[T_d(c)@0]")) {
	externalNogoods = externalNogoods2;
} else {
	externalNogoods = externalNogoods1;
}
```
Then, this is the input program for Alpha:
```
d(c).
b(X) :- aux_r_10000000_2a(a,X), d(X).
a(X) :- aux_r_10000000_2a(b,X), d(X).
a(c) :- g.
b(c) :- g.
```
After 9ca29c0 has been undone now, the error does not occur anymore, but no answer sets are returned for this program (as expected). Besides that, I discovered a theoretical problem concerning atoms that are missing in the negative extension after `assignUnassignedToFalse()` is called, but which the external atom may depend on. Currently, I'm trying to find a theoretical solution. @flowlo , in the meantime we can merge since there is probably no quick fix for this.

PR #15, opened 2017-01-02
Alphahex rule import

tobiaskaminski:
I've adapted the `HexBridge` such that the nogoods imported from Hex are not added directly anymore, but are transformed into nogoods representing the corresponding rules in Alpha. First, I had added a second method to the `Bridge` interface for importing new ground rules. Now, the transformation is made directly inside `getNoGoods`, so that only minimal changes are required in the other classes. Only `RULE_BODIES_PREDICATE` in `NaiveGrounder` has to be _public_ and additional arguments need to be passed to `getNoGoods` for generating new choice atoms. So, I think the new version is better; the the previous one is contained in 111d9f5.


flowlo:
Allowing bridges to generate choices is something I stumbled upon as well and I think we need it. However exposing constants from a concrete grounder implementation and relying on that in a bridge introduces an unwanted dependency. I'll propose some changes.

PR #17, opened 2017-01-05
Improvements to BerkMin

rtaupe:
- ignore atoms that are no choice points
- bugfixes

### DIFF
--- a/src/main/java/at/ac/tuwien/kr/alpha/solver/heuristics/BerkMin.java
+++ b/src/main/java/at/ac/tuwien/kr/alpha/solver/heuristics/BerkMin.java
@@ -64,17 +64,21 @@ public class BerkMin implements BranchingHeuristic {
 	private double decayFactor;
 	private int stepsSinceLastDecay;
 	private Random rand = new Random();
+	private Predicate<? super Integer> isAtomChoicePoint;
 	private Predicate<? super Integer> isAtomActiveChoicePoint;
 
-	public BerkMin(Assignment assignment, Predicate<? super Integer> isAtomActiveChoicePoint, int decayAge, double decayFactor) {
+	public BerkMin(Assignment assignment, Predicate<? super Integer> isAtomChoicePoint,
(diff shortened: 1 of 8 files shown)


src/main/java/at/ac/tuwien/kr/alpha/solver/heuristics/BerkMin.java
```diff
@@ -189,11 +193,24 @@ private void pushToStack(NoGood noGood) {
...
+		// 1. count everything and from time to time do a garbage collection (i.e.
+		// remove atoms that are no choice points)
+		// 2. make check cheaper, e.g. by using dedicated atom IDs (e.g. even
+		// integers for rule bodies, uneven for other atoms)
 	}
 	
 	private void incrementSignCounter(Integer literal) {
-		signCounters.compute(atomOf(literal), (k, v) -> (v == null ? DEFAULT_SIGN_COUNTER : v) + 1);
+		if (isAtomChoicePoint.test(atomOf(literal))) {
+			signCounters.compute(literal, (k, v) -> (v == null ? DEFAULT_SIGN_COUNTER : v) + 1);
```
flowlo: This looks dangerous. Previously, you accessed `signCounters` by unwrapping the atom first (`atomOf(literal)`). Now you are using `literal` directly. Maybe you wanted to do the same as with `incrementActivityCounter`?

rtaupe: No, this is a correct bugfix, see d9bf7b6715a6e5173696b65661e0794b60f332f0.
In `incrementActivityCounter`, we are counting the activity of _atoms_, no matter whether they are positive or negative. In `incrementSignCounter`, however, we count how often an atom occurs positively in nogoods and how often it occurs negatively, therefore we associate the counters with _literals_.
But thank you for reviewing so attentively :)

flowlo: Oh, I understand. :sweat_smile:  I even failed to see the very commit highlighting the change...
Sometimes "better safe than sorry" fires back :blush:

Issue #18, opened 2017-01-05
Command-line interface should provide option to choose branching heuristic


rtaupe:
At the moment, DefaultSolver supports two branching heuristics (see 5a50abae61517e6a5b92ce655cabb784d1dca418), a number which might even increase in the future. Therefore, the command-line interface should provide an option to choose the heuristic(s).

AntoniusW:
This is a good idea. In the future there probably will be several more options required, so we should introduce a common style for setting heuristics-related parameters.

One such parameter, I would like to see, is to turn off any uses of random() inside the heuristics, so that the solving process is completely deterministic. This is extremely helpful during debugging.

flowlo:
I fully agree to the above.

Another attempt to tackle randomness is to have a well-defined source of randomness in our code base (we should have this independently of any CLI params). This would allow us to set the seed for randomness at a single place. A command line option to "disable" randomness would then just use a fixed seed.

This way no code that "uses randomness" needs to change, but we could "disable randomness" very easily.

PR #20, opened 2017-01-05
Get randomness back under control

flowlo:
This addresses the discussion in #18 

 * Single source of randomness, trickles down from `Main`
 * Deterministic and random tests for `DefaultSolver`

### DIFF
--- a/src/main/java/at/ac/tuwien/kr/alpha/Main.java
+++ b/src/main/java/at/ac/tuwien/kr/alpha/Main.java
@@ -24,6 +24,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.util.Arrays;
 import java.util.HashSet;
+import java.util.Random;
 import java.util.Set;
 import java.util.stream.Stream;
 
@@ -40,6 +41,7 @@ public class Main {
 	private static final String OPT_SOLVER = "solver";
 	private static final String OPT_FILTER = "filter";
 	private static final String OPT_SORT = "sort";
+	private static final String OPT_DETERMINISTIC = "deterministic";
 
 	private static final String DEFAULT_GROUNDER = "naive";
 	private static final String DEFAULT_SOLVER = "default";
@@ -84,6 +86,9 @@ public class Main {
 
 		Option sortOption = new Option("s", OPT_SORT, false, "sort answer sets");
 		options.addOption(sortOption);
+
+		Option deterministicOption = new Option("d", OPT_DETERMINISTIC, false, "disable randomness");
+		options.addOption(deterministicOption);
 
 		try {
 			commandLine = new DefaultParser().parse(options, args);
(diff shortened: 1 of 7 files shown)


src/main/java/at/ac/tuwien/kr/alpha/Main.java
```diff
@@ -81,6 +83,9 @@ public static void main(String[] args) {
 		filterOption.setValueSeparator(',');
 		options.addOption(filterOption);
 
+		Option deterministicOption = new Option("d", OPT_DETERMINISTIC, true, "disable randomness");
```
AntoniusW: I think the third parameter should be `false` to indicate that the option takes no argument?

src/main/java/at/ac/tuwien/kr/alpha/solver/heuristics/BerkMin.java
```diff
@@ -56,25 +56,31 @@
...
+		this.rand = random;
+	}
+
+	public BerkMin(Assignment assignment, Predicate<? super Integer> isAtomActiveChoicePoint, Random random) {
+		this(assignment, isAtomActiveChoicePoint, DEFAULT_DECAY_AGE, DEFAULT_DECAY_FACTOR, random);
 	}
 
 	public BerkMin(Assignment assignment, Predicate<? super Integer> isAtomActiveChoicePoint) {
-		this(assignment, isAtomActiveChoicePoint, DEFAULT_DECAY_AGE, DEFAULT_DECAY_FACTOR);
+		this(assignment, isAtomActiveChoicePoint, DEFAULT_DECAY_AGE, DEFAULT_DECAY_FACTOR, new Random());
```
AntoniusW: Isn't the `new Random() `violating the idea of random passed down from top?

src/main/java/at/ac/tuwien/kr/alpha/Main.java
flowlo: Yep. Will fix in a second.

src/main/java/at/ac/tuwien/kr/alpha/solver/heuristics/BerkMin.java
flowlo: I wanted to keep the "all default" constructor, that will construct it's own `Random`. This will be used for example in `BerkMinTest`. This way the constructor stays compatible with existing code. I agree that always requiring some source of randomness to be passed would be more strict. However you like.

AntoniusW: Personally, I like the strict version more (i.e., delete the constructor and create own `Random` in the test). But I would also like to hear @rtaupe's opinion on the whole idea.

flowlo:
Note that we could also allow the user to pass a concrete number to be used as seed (thinking of bugs that only occur with certain seeds). This would be straightforward to implement, but as long as we don't really need it, I think that just "switching off" randomness by using a seed of zero is sufficient.

src/main/java/at/ac/tuwien/kr/alpha/solver/heuristics/BerkMin.java
flowlo: Sure!

rtaupe: I am with @AntoniusW on this, we should be consequent and instantiate `Random` at only one place.

rtaupe: Except in test cases where we don´t have access to the central rand.

PR #21, opened 2017-01-09
Adding some pigeon hole test cases.

rtaupe:
Note that I actually intended to show by the introduction of these test cases that Alpha is currently unable to solve even the simplest pigeon hole test cases within a few seconds. Then however I noticed that this is only the case if the log level is DEBUG, which shows that the debug mode makes Alpha _really_ slow.
The issue persists, however, with test cases like `test4Pigeons3Holes` or `test3Pigeons4Holes`, which are currently unsolvable also on log level INFO.
Alpha needs to get faster!


src/test/java/at/ac/tuwien/kr/alpha/solver/AbstractSolverTest.java
```diff
@@ -514,4 +512,97 @@ public void simpleNoPropagation() throws IOException {
 		Set<AnswerSet> answerSets = solver.collectSet();
 		assertEquals(expected, answerSets);
 	}
+
```
AntoniusW: Those tests are a good addition. Since they come with specific helper and construction methods and these are all pigeon hole specific, I suggest do move them to their own class.

rtaupe: Good idea. Is now done.

flowlo:
After 2ddea5c957c134f412982a86d0f4521ac885a635 the tests are not runnable tests anymore (class is abstract). There are (at least) two possible solutions:

 * Make the class concrete and instantiate some solver (this could be done in subclasses as well): #26 
 * Restructure our tests (I'd prefer that): #25

rtaupe:
@flowlo Thanks! :)

Issue #31, opened 2017-01-18
Command-line interface should be able to accept multiple input files


rtaupe:
In ASP it is common to provide problem encoding and instance in two separate files. Therefore, Alpha should be able to accept several input files via the -i command-line argument.

AntoniusW:
Functionality is now present in master.

PR #32, opened 2017-01-18
Accept multiple input files

flowlo:
This is a quick solution to #31 

In principle it does what it needs to do. What we will need in the future is referencing the file name in case of parsing errors. Currently files are concatenated and the line numbers that ANTLR error messages refer to are probably not consistent with file numbers of actual files.

So I'd say it's a step in the right direction, but we'll need to follow up on the issue.


flowlo:
With d1a5c92 our error messages also have file names associated with them, yay! Also, I decided to get rid of our `stream` util (that would wrap a stream around a string) because ANTLR has `ANTLRInputStream` that we can directly use.

PR #38, opened 2017-01-19
Refactoring

flowlo:
1. Use `Atom` where possible.
 1. `Atom` now offers `getPredicate()`, `getTerms()` and `isInternal()` and it `extends Comparable<Atom>`.
 1. Instead of `SubstitutionUtil` we now have
    * `Term.substitute(Substitution)` which generates a substituted `Term` from a `Term`
    * `Atom.substitute()` which generates a substituted `Atom` from an `Atom`
 1. Separate `Atom` implementations for choices and rules to get rid of magic predicate names as far as possible.
 1. Depend on [Google Guava](https://github.com/google/guava).
    * It has interners and some more useful utilities that we might want to use in the future.
 1. Remove `AtomId`.
 1. Remove `ReteGrounder` (was a stub).

`NaiveGrounder` now contains way less casting and is ~100 lines shorter (mostly because `Substitution`) now is a standalone class. By adding `Atom.substitute()` we allow all implementations of `Atom` to supply their own implementation, which is less centralized than `SubstitutionUtil` but more object-oriented.


src/main/java/at/ac/tuwien/kr/alpha/common/BuiltinAtom.java
```diff
@@ -1,117 +1,164 @@
...
 public class BuiltinAtom implements Atom {
-	public final Term left;
-	public final Term right;
-	public final ParsedBuiltinAtom.BINOP binop;
+	private final List<Term> terms;
+	private final ParsedBuiltinAtom.BINOP binop;
+
+	protected BuiltinAtom(ParsedBuiltinAtom.BINOP binop, List<Term> terms) {
+		this.binop = binop;
+		this.terms = terms;
```
AntoniusW: Please add a check here to guarantee that a BuiltinAtom consists of only two terms. (Throw RuntimeException to abort.)

src/main/java/at/ac/tuwien/kr/alpha/common/FunctionTerm.java
```diff
@@ -1,67 +1,96 @@
...
-	public static FunctionTerm getFunctionTerm(String functionSymbol, List<Term> termList) {
-		return getFunctionTerm(TermSymbol.getInstance(functionSymbol), termList);
+	public Symbol getSymbol() {
+		return symbol;
 	}
 
-	public static FunctionTerm getFunctionTerm(String functionSymbol, Term... terms) {
-		return getFunctionTerm(TermSymbol.getInstance(functionSymbol), Arrays.asList(terms));
+	public boolean isSimilar(FunctionTerm other) {
+		return symbol.equals(other.symbol);
```
AntoniusW: isSimilar is a bad naming, it checks equality of function symbols and no similarity. Maybe call it `equalFunctionSymbol` ?

flowlo: I removed this in favor of an explicit check in `Substitution`: `a.getSymbol().equals(b.getSymbol())`. `isSimilar()` was only called once anyways.

src/main/java/at/ac/tuwien/kr/alpha/grounder/Substitution.java
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	public Substitution(Substitution clone) {
+		this(new HashMap<>(clone.substitution));
+	}
+
+	public void replaceSubstitution(Substitution other) {
+		substitution.clear();
+		substitution.putAll(other.substitution);
+	}
```
AntoniusW: Is there a reason to clone the substitution here?

flowlo: The main difference [to the previous version](https://github.com/AntoniusW/Alpha/pull/38/files/55317968094475905c0194fa76313f23fbd5a62a#diff-3bb08e391a45e7c42bba1e6471511498L665) is that `this.substitution` is final now, so the reference cannot be be changed. I will revert that.

.idea/modules/alpha_main.iml
```diff
@@ -21,6 +21,7 @@
     <orderEntry type="library" name="Gradle: org.apache.commons:commons-collections4:4.1" level="project" />
     <orderEntry type="library" name="Gradle: org.apache.commons:commons-lang3:3.4" level="project" />
     <orderEntry type="library" name="Gradle: ch.qos.logback:logback-classic:1.1.7" level="project" />
+    <orderEntry type="library" name="Gradle: com.google.guava:guava:20.0" level="project" />
```
flowlo: Google Guava is licensed under the Apache License 2.0, which is (conceptually) derived from MIT and BSD. There's some discussion about it on [SE](http://softwareengineering.stackexchange.com/questions/40561).

The APL-2.0 says:

> For the purposes of this License, Derivative Works shall not include works that remain separable from, or merely link (or bind by name) to the interfaces of, the Work and Derivative Works thereof.

Alpha remains separable from APL, as we can (theoretically) just use different implementations of their interfaces.

With this definition I would say that Alpha is not a Derivative Work of Google Guava, so we don't have to do anything. If you for some reason would see Alpha as a Derivative Work, we would have to include the original license (APL-2.0) and copyright of Google Guava. Google Guava does not include a NOTICE file, so nothing to do there.

By the way: We also depend on ANTLR (Licensed under BSD) and Apache Commons (licensed under APL-2.0) in a similar way.

PR #53, opened 2017-04-27
Modified watches

AntoniusW:


src/main/java/at/ac/tuwien/kr/alpha/solver/Assignment.java
```diff
@@ -106,6 +108,39 @@ default boolean guess(int atom, boolean value) {
 	int getDecisionLevel();
 
 	/**
+	 * Helper for debugging: prints the given NoGood with its assignment inlined.
+	 * @param noGood the nogood to print.
+	 * @return a string of the nogood with the current assignment (inlined).
+	 */
+	default String printNoGoodAssignment(NoGood noGood) {
```
lorenzleutgeb: This does not actually print. It just returns a string.

src/main/java/at/ac/tuwien/kr/alpha/solver/ThriceTruth.java
```diff
@@ -57,6 +57,13 @@ public static ThriceTruth valueOf(boolean value) {
 	}
 
 	/**
+	 * @return true if this is MBT.
+	 */
+	public boolean isMBT() {
```
lorenzleutgeb: Why not just let users of this type check `x == MBT`? That's absolutely reasonable and clear. What is the reasoning behind adding `isMBT`?

AntoniusW: My intention was to have a shorthand for MBT.equals(truth), but I think, we could replace this and all ThriceTruth.MBT/TRUE/FALSE.equals with simple == because ThriceTruth is an enum. Correct?

src/main/java/at/ac/tuwien/kr/alpha/solver/Assignment.java
AntoniusW: Well, the IDE debugger prints it. ;-)
What would you suggest to name it?

PR #55, opened 2017-04-29
Moving to sudo-enabled Travis CI (to avoid test failure).

AntoniusW:
For some unknown reason, Travis CI kills the build while running all tests. Enabling sudo seems to avoid this issue.


lorenzleutgeb:
This is hilarious.

PR #60, opened 2017-09-11
Term intervals

AntoniusW:
Supporting integer intervals as terms, which allows facts like `foo(1..5).` similar as Clingo.


lorenzleutgeb:
I really like the idea of allowing term intervals. However, I have the feeling that this implementation stretches too far into our internals. My suggestion would be to resolve `foo(1..5)` to `foo(1)` through `foo(5)` before the program is even translated into our internal classes such as the `Atom` family.

What is the justification for having `IntervalAtom` that introduces lots of complexity versus doing a rather simple re-write step that expands intervals in the parser stage?

Pollution of what should be clear concepts like `Atom` worry me. We now have real atoms (e.g. `BasicAtom`), symbolic atoms (`RuleAtom` and `ChoiceAtom`) and a special thingy for builtins.

AntoniusW:
I fully agree with you that less intrusive term intervals are better.
For `foo(1..5)` this would work, but term intervals (according to the Clingo/gringo user manual) are allowed at any position where a term might occur. So we can have term intervals in rules like
`p(X, 1..X) :- dom(X), X >= 2.`
where it is no longer possible to compile them away without computing the full grounding of `dom(X)` first. Clearly, we do not want this, so the grounder must be able to expand term intervals lazily during lazy grounding. Therefore some representation of those terms is needed. The current approach to enable interval expansion during lazy grounding is to rewrite term intervals into a new variable plus a special `IntervalAtom` which then produces substitutions for the introduced variable. So the above rule becomes
`p(X, _I1) :- dom(X), _intervalAtom(_I1, 1, X), X >=2.`
By that, we can at least leave the rest of the grounder unchanged, intervals simply generate new variable substitutions as the grounder is handling them already.

The only improvement I see, would be to make `IntervalAtom` a subclass of `BuiltinAtom`, but this is semantically not fully correct and it creates trouble with ordering the atoms in the rule body such that variables are bound when used by builtin atoms. The latter issue actually is a larger one that needs its own solution (e.g., currently `X = 2` is considered as not binding `X`), but this solution requires a larger analysis of the order in which variables are bound in a rule. This issue of binding also touches several other problems which we are going to address in the coming months. For the moment, I'd prefer the less-clean solution, since we have to come up with a clean one in a _more general_ setting anyways.

lorenzleutgeb:
I see. I also had in mind that you probably would want to expand the interval only in the grounder ...

lorenzleutgeb:
I changed `IntervalTerm` so that it checks for whether it is ground at construction time, assuming that `isGround()` would be called at least once for most instances. Also, if things go wrong, this will fail earlier (at construction time of the object, not when calling a getter).

In `NonGroundRule` you introduce some static methods, where dynamic ones, such as `FunctionTerm.rewrite` would be possible, and maybe easier to read (less casts). Your call.

AntoniusW:
Indeed, the rewriting is quit ad-hoc. There surely will be more of that in the future, but I am not sure where to put this. Moving it into FunctionTerm would probably scatter the rewriting too much (since I forsee that there are multiple forms of rewriting ahead). Maybe `grounder.transformation` would be the right place for that? There, however, we can only work with the parse tree, hence we cannot directly insert `IntervalAtom` but only (ordinary) atoms which then later have to be recognized again or we create a new `ParsedIntervalAtom`. Another solution would be to rewrite the internally generated structures (i.e., `NonGroundRule` similar as it is now). Any ideas on the whole issue of where to put rewritings?

lorenzleutgeb:
Hmm, I'd say we agree that `NonGroundRule` should have less logic and also that we do not want to scatter rewriting. What about introducing rewriters that work on `NonGroundRules` and generate ground instances? They'd have to be composable such that multiple rewriters can be combined.

AntoniusW:
Introducing stackable rewriters that work on `NonGroundRules` sounds like a good idea. Then, we could also move some of the current logic in `NonGroundRule` to some other place (e.g., for checking safety or ideal atom ordering). These rewriters, I think, should output `NonGroundRules` again and not yet ground instances, since this is part of the lazy grounder (`NaiveGrounder` currently). Side note: `NonGroundRules` may be ground, so maybe we should rename that class to `AlphaRule` or something similar. I will try some refactoring in the near future, thanks for the input!

lorenzleutgeb:
`AlphaRule` is kind of redundant ... However:

```java
@FunctionalInterface
interface Rewriter {
    Rule rewrite(Rule rule)
)
```

:)

PR #61, opened 2017-09-22
Introduce Evaluable interface for Predicates

lorenzleutgeb:
I'd like to bring up a question/suggestion by example. Note that the set of changes in this PR is rough, and primarily meant to illustrate for discussion.

Ever since `BuiltinAtom` was introduced I was wondering whe *built-in* qualifier is associated with the atom, not the predicate. Currently, builtin atoms cover (in-)equality comparison for variables and classical ordering of integers. From my point of view, we are defining/building-in the behaviour of these comparison operators, and they are predicates. The truth value of corresponding atoms thus can be computed by evaluating the predicate given the terms of the atom.

My guess is that the association was taken from [the ASP-Core-2 paper](https://www.mat.unical.it/aspcomp2013/files/ASP-CORE-2.03c.pdf), which also talks about *built-in atoms*.

In this PR I exemplified my thinking: I introduced a new type called `Evaluable` which is used to mark predicates that can be evaluated directly, e.g. without considering any rules in the program. I implemented `TotalOrder implements Evaluable, Predicate` as a replacement for the binary operations from `BuiltinAtom`.

The reason why I am interested in predicates that can be evaluated is because this is where I would like to start experimentation to interface with "ordinary" Java programs, e.g. I'd like to provide a way to load `Evaluable` predicates from "the outside":

```java
class Graph {
    @Predicate
    public boolean connected(Node a, Node b) {
        ...
    }
}
```

With that, I think I could get a clearer picture of how we'd go about integrating external atoms, similar to what HEX does.


AntoniusW:
First of all, I like the idea of directly interacting with Java or other pluggable external evaluations very much. The observation about `BuiltinAtoms` actually being built-in predicates is also correct, the naming stems from the ASP Core standard and we may change it.

The above discussion, however, does not yet consider the case that external evaluations may generate output, as well as the fact that some of the `BuiltinAtoms` also should be able to generate variable bindings (which they currently cannot). In the end, a rule like the following should be perfectly fine:
`succ(X,Y) :- dom(X), Y = X +1.`
So, the built-in equality may bind variables (given that there is a sole variable on one side of the equation; an equality like `Y-X-1 = 0` for example does not generate bindings). I think that external atoms /  `Evaluable` predicates should also be able to generate variable bindings. I think, we should introduce some notion to indicate which atom positions can generate bindings, in order to check whether a rule is safe and to have a fixed evaluation ordering of the atoms that guarantees that all necessary variable bindings are generated.

By the way, I think the condition of when an atom can be evaluated directly depends not on any other rule(s) but is a matter of the interpretation of the underlying logic. We can evaluate builtin equality, because the truth value of equality is nothing that can be changed by an interpretation or an answer set. The builtin `3 +1 = 4` shall be true in all interpretations, because we want the interpretation of `=` to be fixed and not something the interpretation may pick. This has influences on external evaluations as follows: if the external evaluation depends (or may depend) on the interpretation or answer set, then the evaluation cannot be done upfront but must be re-done every time the assignment changes. In HEX this is the case, since external atoms in HEX may receive predicate input, i.e., they receive the extension of a predicate (in the current assignment) as input and their truth value (and output) depends on this extension.

If we restrict external evaluations to receiving terms as input, then the evaluation is fixed and not depending on any interpretation (remember, we user Herbrand interpretations where each term is interpreted with itself and thus this is fixed). Hence it can be done as you intended it in this PR. I suggest to follow this way, but it is clear that this approach cannot cover all that is possible with HEX. The advantage of this approach is, that external evaluations are possible, they are relatively simple, and their evaluation can be done quite fast. Since using HEX and Alpha is possible anyways (by using the respective branches of these systems), I see no drawbacks in adding "light" external atoms.

lorenzleutgeb:
OK, I think we're on the same page. Restricting input to terms will save lots of headaches and should still be a useful feature on its own.

I just pushed some more (still incomplete) changes. I extended the grammar for external atoms. Also, my plan is to resolve implementations of predicates through reflection, i.e. scan for annotated methods as sketched in my first post. Alpha would then keep maintain a function table.

src/main/antlr/at/ac/tuwien/kr/alpha/antlr/ASPCore2.g4
```diff
@@ -42,12 +42,14 @@ weight_at_level : term (AT term)? (COMMA terms)?;
...
 naf_literals : naf_literal (COMMA naf_literals)?;
 
-naf_literal : NAF? (classical_literal | builtin_atom);
+naf_literal : NAF? (external_atom | classical_literal | builtin_atom);
 
 classical_literal : MINUS? ID (PAREN_OPEN terms PAREN_CLOSE)?;
 
 builtin_atom : term binop term;
 
+external_atom : MINUS? AMPERSAND ID (SQUARE_OPEN input = terms SQUARE_CLOSE)? (PAREN_OPEN output = terms PAREN_CLOSE)?;
```
AntoniusW: The first parentheses for input cannot be optional, as `&ext(foo)` is ambigous, i.e., `foo` could be input or output. I would also drop the optional MINUS -- this is not supported anyway.

lorenzleutgeb: I'd agree if input and output list would use the same parentheses, though output uses the `[]` pair while input uses `()`, inspired by [DLV-HEX (Section 2)](http://www.kr.tuwien.ac.at/staff/eiter/et-archive/iswc05-dlvhex.pdf).

Will remove `MINUS?`.

AntoniusW: Yes, the `MINUS?` probably is not needed. Sorry, I twice saw the wrong parentheses.

AntoniusW:
I am not sure if scanning via reflection is the right approach, since this must be done every time and reflection is usually quite slow. What I have seen is a default location for plugins and a command-line option to specify precisely which plugins to load and another option to specify where to look for additional plugins.
It also might be possible that there are two plugins that introduce the same external atom. If they are not used at the same time, they should be both okay.
I think there should be some kind of configuration (command line, special file, file system location), but I am not sure which would be preferable. Any ideas or experience on that?
I guess a command line option to specify which plugins should be loaded is better than the others, because it allows easy selection and benchmarking. This option then could be the class names of those plugins that are to be loaded (okay, reflection again, but this time limited to only the actually used plugins).

On another note: why change from constants to objects, done in commit 34ee0b7?

lorenzleutgeb:
Indeed, scanning for annotations is slow, but it is a widely used practice in enterprise code (consider [this cheat sheet for the Spring framework](http://files.zeroturnaround.com/pdf/zt_spring_annotations_cheat_sheet.pdf) which uses annotation metadata to power its IoC container, or [for Hibernate, a Java Persistence API implementation](http://www.cheat-sheets.org/saved-copy/rc032-010d-hibernate_search_0_1.pdf#page=5), or [for JPA itself](http://www.oracle.com/technetwork/middleware/ias/toplink-jpa-annotations-096251.html#IndexOfAnnotations)). Annotations have largely replaced configuration files in these ecosystems. My motivation here was to provide people coming from an application development perspective to coding a way to integrate with Alpha that they already are familiar with. But yeah sure, just passing the names of methods via CLI or a config file would work perfectly fine and is easy to implement.

Could you please elaborate what you mean by a "plugin" in this context?

Note that scanning for annotations is optional, the consumer of the API might also directly `register` a `Method`. I don't see any faster way to obtain a reference to a method, except as exercised in `NativeEvaluablePredicate` which wraps `java.util.function.Predicate`.
Just to clarify on "every time": Annotations have to be scanned before the solver starts. During solving, no scanning takes place.

Regarding name clashes: Yes, you are right, the current implementation suffers from that, as names are unstructured and shared globally. Methods are always inside a class, which are always inside a package (even the default package is a package). So we might as well store this fully qualified name, and would get to re-use the naming scheme of Java.

I changed constants to be implemented as Java objects, because I thought that would open up new possibilities of integrating with Java applications: Up to now, all programs were files that were parsed into our internal data structure. It only makes sense that all terms are strings, because that's the only structure we can easily represent in text files¹. If we relax the constraint that programs are read from files, but might instead be [constructed by application programs](https://github.com/alpha-asp/Alpha/pull/61/commits/34ee0b7e57ddc51a9c8db6051035de7e5c684b09#diff-97b604b0a024beb1b5d1c52adb1564efR81), then questioning why terms are strings seems natural to me.

Consider the following:

```
person(joe).

% Some choice that either selects or does not select each person.

% Normally we would probably say that Joe is a man:
% man(joe).

% But nothing holds us back from using an external predicate.
:- selected(X), &man(X).
```

We are talking about some object Joe, and represent it by the string `joe`. But what if we already have a different, structured representation of Joe, for example a Java object of type `Person`. It would be cool if the external predicate could directly look at a Java object, and not a `String`, which then maps to the Java object.

TBH I did not really think through all details (I hacked all recent changes on an overnight bus trip from Dresden to Vienna) but I think that it is not that important to the ASP system what terms actually are. The change I made is quite minute. Also, note that in Java all strings are objects of type `String`, so in the end this is "just" a generalization.

Programs that are read from files, will just have `String`s as their constants (the change is backward compatible), while others that are instantiated from an application program might have other objects as constants. From a performance perspective, this should be negligible, but I will benchmark once the code gets more stable. Also, I plan on using generics to avoid some bad casts at runtime.

How do you feel about constants as objects?

src/main/java/at/ac/tuwien/kr/alpha/common/predicates/ExternalEvaluable.java
```diff
@@ -37,7 +38,11 @@ public boolean evaluate(List<Term> terms, Substitution substitution) {
 				it = it.substitute(substitution);
 			}
 
-			arguments[i] = it;
+			if (!(it instanceof ConstantTerm)) {
+				throw new IllegalArgumentException("how did i get here?");
```
AntoniusW: I guess execution of this branch indicates a bug. Since users might see this first, the message should be more informative like "[what happened]. Should not happen."

src/main/java/at/ac/tuwien/kr/alpha/common/predicates/Evaluable.java
```diff
@@ -0,0 +1,11 @@
+package at.ac.tuwien.kr.alpha.common.predicates;
+
+import at.ac.tuwien.kr.alpha.common.terms.Term;
+import at.ac.tuwien.kr.alpha.grounder.Substitution;
+
+import java.util.List;
+
+@FunctionalInterface
+public interface Evaluable {
+	boolean evaluate(List<Term> terms, Substitution substitution);
```
AntoniusW: Similar as External, the Evaluable should be able to return variable bindings. According to the standard (but not yet realized in Alpha), `=` may bind a single variable on either side to the value of the other side.

src/main/java/at/ac/tuwien/kr/alpha/common/predicates/External.java
```diff
@@ -0,0 +1,10 @@
+package at.ac.tuwien.kr.alpha.common.predicates;
+
+import at.ac.tuwien.kr.alpha.common.terms.Term;
+import at.ac.tuwien.kr.alpha.grounder.Substitution;
+
+import java.util.List;
+
+public interface External {
+	int evaluate(List<Term> input, List<Term> output, Substitution substitution);
```
AntoniusW: It is not clear to me why there is `input`, `output` and a `substitution`. On the one hand, the substitution may be used to return variable bindings, on the other hand it maybe is not good to expose the substitution to any custom code (which might try to modify it).

I think it would be best to have only ground input and ground output. By that, the substitution must be handled outside by the grounder, which I guess is the right place anyways, and the `External` is not at all concerned with variables.

src/main/java/at/ac/tuwien/kr/alpha/common/Symbol.java
```diff
@@ -9,22 +9,25 @@
...
+		this.id = id;
 	}
 
 	public static Symbol getInstance(String symbol) {
 		return getInstance(symbol, 0);
 	}
 
 	public static Symbol getInstance(String symbol, int arity) {
-		return INTERNER.intern(new Symbol(symbol, arity));
+		return INTERNER.intern(new Symbol(symbol, arity, counter++));
```
AntoniusW: Does this mean that `Symbol.getInstance("a",0) != Symbol.getInstance("a",0)` ? It looks wrong to me, why is this okay?

AntoniusW:
I see you had much more in mind than what I saw. Thank you for all the explanations.

Regarding the configuration of external atoms: last time I worked with Spring it had me follow a method invocation routed through 3 or 5 XML files and I have not enough experience with annotations to give a good judgement of drawbacks and advantages. I see no strong argument against annotations, except if it increases the startup time significantly. For benchmarking, it might be useful to have an easy way to activate, deactivate, and configure external atoms (probably via CLI). By "plugin" I meant the code behind an external atom (terminology from HEX).

Using fully qualified names for external atoms probably is overkill. Although this touches another longstanding open issue of logic programming: modules and scoping. For now, there is no simple formalism to use in such cases. I would therefore suggest to simply disallow the simultaneous use of plugins that have the same external atom.

Regarding constants: ohh, I like your idea! Now I also understand that test case. In general, I follow your explanation, but I want to caution the identification of terms and strings. Formally, terms are just elements of a certain set and there is no relation to strings. Of course, we have a string representation of every term and the representation is such that two terms are equal iff their string representations are equal. Seems like nitpicking but I think it actually makes a difference, if terms can be objects. So, our code should rely on object.equals for term-objects.

lorenzleutgeb:
Thank you for your comments, they all are spot on and I will address them. But please bear with me for a little longer, as I consider the set of changes in this branch highly experimental and they will "solidify" over time. What I wanted to achieve now is to get a proof of concept working as fast as possible, with all pros and cons :)

Re annotations:
Startup time will only increase significantly if someone calls `scan`. And they'd probably only do if they have annotated at least one method. For uses where startup time is critical and should be predictable, they should use `register`, or CLI arguments, or a config file (the last two are not implemented yet, of course), as in these cases, zero annotation scanning is done.

Re names of external atoms:
Yes, I decided to not go into details with structured names and uniqueness of names as there are issues with that in general. Agreed, fully qualified names for predicates would be overkill, but my point was that *if necessary* we can easily support FQNs for external atoms derived from Java methods.
Short term solution: Throw an exception as soon as a name clashes and refuse operation.

Re constants as objects:
"terms are just elements of a certain set" Yes! :D That's exactly what gave me confidence in trying out the change. I just define my certain set as the set of Java objects (which luckily happens to include the set of all strings, for that matter) :) Consequently I agree, `Object.equals` should be relied on.

Some other systems (DLV) guarantee an ordering of terms, which is quite handy sometimes, so we might consider restricting ourselves to the set of `Object`s that implement `Comparable`...

AntoniusW:
No pressure, those changes are interesting, they run quite deep, and need time. Sorry if my comments made another impression.

Re annotations: I see, so scanning is no default. That's good.

Re names of external atoms: the short term solution is reasonable and I support it.

Re term ordering: this is another issue. The ASP Core standard actually defines a certain term ordering, which we do not implement. But we should. Java objects could then be added at one end of the standard's order.

AntoniusW:
I am not sure where you are heading with the modifications of the parser. I guess you want to parse ASP programs as well as answer sets?

You can do this directly in `ASPCore2.g4` by just adding a new rule
`answer_set : CURLY_OPEN classical_literal (COMMA classical_literal)* CURLY_CLOSE;`
Whenever you want to parse an answer-set you call then `parser.answer_set()` instead of `parser.program()` and you get the parse tree of the answer-set (cf. `Main.parseVisit`). This, of course, requires a new method in the `ParseTreeVisitor` where the parsed answer set is transformed into an actual one.

The current approach towards this seems to include some refactoring of the input to the parser generator. I do not like this for the following reasons. First, the code in `ASPCore2.g4` currently can be directly compared with the specification in the actual standard (it is almost one-by-one). Second, it is not a lot of code, less than 130 lines, and by splitting it into multiple files one has to chase more than by simply scrolling a few lines. Third, it seems you actually are introducing duplicate code now (second ParseTreeVisitor shares a lot with the already-existing one, new Parser class and subclasses).

I am not sure if the above `answer_set` rule for the parser generator is enough since I do not know your actual goals, but I would like to not scatter the parser code more than necessary.

lorenzleutgeb:
Yes, I want to have a parser for answer sets that we can use instead of `AnswerSetBuilder` in our tests and also to compare the output of Alpha with answer sets from a file (not just doing string comparison).

The reason why the parser for answer sets is separate from the ASPCore2 parser is twofold. Firstly, because I assume that we only need to parse answer sets when testing. Therefore I put it into the "test" source set, and it will not bloat Alpha when distributed. Secondly, I see where you are going with the idea of adding rules to adding a rule to the ASPCore2 parser to do the job, but one rule alone does not suffice. For answer sets I'd like to exclude, variables, arithmetic and intervals. Introducing this distinction (ground terms vs. non-ground terms) to the ASPCore2 grammar would be possible, but would probably at the same time destroy the benefit of it closely mirroring the spec paper.

I added a separate parser for terms only, because the regular expression based parser used in `BasicAnswerSet.Builder` did not capture the difference between symbol constants and string constants, and I did not see a simple way to extend it. Probably I should remove the second parser, and just use the answer set parser for tests that require parsing of terms, or require construction of function terms using `FunctionTerm`, not just strings.

AntoniusW:
Ah, now I better understand your motivations. I think it is a good idea to have an answer set parser (also for files), because it has good use even beyond testing. Sometimes when writing an ASP program one wants to debug the ASP program, because it yields other answer sets than one would expect. There are cases when it is useful to have an easy method of comparing desired output with actual output, e.g. to figure out which part of the ASP program is wrong, or which answer sets are missing. Even though I have no ready idea of how this should be used, I think one can expect it to be used by users of the solver and not just for our internal testing.

Regarding the second argument: checking that the given answer set is indeed free of variables, intervals, and builtin atoms is actually quite easy once we got the parsing result. Simply iterate over all atoms and check whether they are ground and free of interval terms. For both checks, the code is already present (the latter currently is in `FactIntervalEvaluator.functionTermContainsIntervals`). Builtin atoms fail the parsing since they are not `classical_literal`. So these checks should be doable in about 10 lines of code using the above parser rule. To me, this feels like much less bloat.

lorenzleutgeb:
OK.

By now I have removed the term parser, and rewrote some of our tests to use an answer set parser. I will now also remove the separate grammar for answer sets and instead re-use the ASPCore2 grammar with an additional check.

PR #62, opened 2017-10-12
Refactoring parsing-related code to remove almost-duplicate types.

AntoniusW:
- ParseTreeVisitor now directly generates Terms, Atoms, Rules, etc. instead of creating intermediate ParsedTerm, ParsedAtom, etc.
- Removing unnecessary ParsedTerm, ParsedAtom, etc.
- Moving ChoiceAtom and RuleAtom from common to grounder packages.
- Adapting all test cases.


AntoniusW:
It seems that the AppVeyor CI needs some missing configuration files. Lorenz?

lorenzleutgeb:
Cool. I wanted to do this for quite some time ... Will take a closer look soon.

src/main/java/at/ac/tuwien/kr/alpha/common/Rule.java
```diff
@@ -0,0 +1,140 @@
...
+	/**
+	 * Returns the n-th atom in the body of this non-ground rule.
+	 * @param atomPosition 0-based position of the body atom.
+	 * @return
+	 */
+	public Literal getBodyLiteral(int atomPosition) {
+		return bodyLiterals.get(atomPosition);
+	}
+
+	public int getNumBodyAtoms() {
```
lorenzleutgeb: I'd go for something that uses Java terminology ("size", "count", "length"), e.g. `getBodySize()`.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/Literal.java
```diff
@@ -0,0 +1,13 @@
+package at.ac.tuwien.kr.alpha.common.atoms;
+
+/**
+ * Copyright (c) 2017, the Alpha Team.
+ */
+public interface Literal extends Atom {
+
+	boolean isNegated();
+
+	default boolean isPositive() {
```
lorenzleutgeb: Please remove `isPositive()` ... There should be only one possibility to get the polarity of a literal. `isPositive()` is redundant in a very obvious way, and it will only make reading our code harder (two variants capture the same concept).

src/main/java/at/ac/tuwien/kr/alpha/grounder/internalatoms/ChoiceAtom.java
```diff
@@ -1,8 +1,9 @@
-package at.ac.tuwien.kr.alpha.common.atoms;
+package at.ac.tuwien.kr.alpha.grounder.internalatoms;
```
lorenzleutgeb: I very much like the separation. "internalatoms" is unusually long ... I think that "atoms" should suffice.

src/main/java/at/ac/tuwien/kr/alpha/grounder/parser/ParseTreeVisitor.java
```diff
@@ -0,0 +1,286 @@
...
+			binop = BuiltinAtom.BINOP.LE;
+		} else if (parsedBinop.GREATER_OR_EQ() != null) {
+			binop = BuiltinAtom.BINOP.GE;
+		} else {
+			throw new RuntimeException("Unknown binop encountered.");
+		}
+		return new BuiltinAtom(binop, termList, isCurrentLiteralNegated);
+	}
+
+	private boolean isCurrentLiteralNegated;
```
lorenzleutgeb: Please move this declaration to the top of the class.

src/main/java/at/ac/tuwien/kr/alpha/grounder/parser/ParseTreeVisitor.java
```diff
@@ -0,0 +1,286 @@
...
+
+		if (ctx.statements() == null) {
+			return Program.EMPTY;
+		}
+		inputProgram = new Program(new ArrayList<>(), new ArrayList<>());
+		visitStatements(ctx.statements());
+		return inputProgram;
+	}
+
+	private Program inputProgram;
```
lorenzleutgeb: Please move this declaration to the top of the class.

src/main/java/at/ac/tuwien/kr/alpha/common/Program.java
```diff
@@ -0,0 +1,35 @@
...
+ * Copyright (c) 2017, the Alpha Team.
+ */
+public class Program {
+	public static final Program EMPTY = new Program(Collections.emptyList(), Collections.emptyList());
+
+	private final List<Rule> rules;
+	private final List<Atom> facts;
+
+	public Program(List<Rule> rules, List<Atom> facts) {
+		this.rules = rules;
```
lorenzleutgeb: IMO, the private members of `Program` should be `final`, with the consequence of a `Program` being immutable. This can be achieved by wrapping the arguments with [`Collections.unmodifiableList`](https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#unmodifiableList-java.util.List-).

This would force anyone to first come up with a set of rules (and facts) and only then instantiate a program, which then can never be changed. It makes some things simpler for us, e.g. no one can add a rule to the program while we are currently solving it (which might otherwise lead to undefined behavior).

In the `ParsedTreeVisitor` you'd probably have to accumulate two separate lists, and construct the `Program` from them.

If you that's too cumbersome, I am happy to make these changes in a follow-up PR.

AntoniusW: I thought about the same problem but decided against making `Program` immutable for the following reasons:
1. Each program transformation would have to create copies (I have plans to add two or three in the next days/weeks). We may have hundreds of thousands of facts, potentially also many rules. Copying seems very unnecessary when just a few rules get added and removed.
2. At Aalto, we have been thinking about lazy instantiating some transformations, e.g., to introduce aggregates. In a ground-and-solve setting these aggregates have to be instantiated during grounding, but in our setting we can postpone that (possibly even until very late in the solving). Not fully sure yet, but in these cases it might be necessary to have a mutable `Program`.
3. If we introduce an outward-facing API, we could just deep-copy the `Program` given to the API. Then a later modification from the outside does not touch the solver's private copy, hence no undefined behaviour.

Maybe 3. is not the best solution to this problem, so I am open to other approaches.

lorenzleutgeb: OK, I agree :)

PR #63, opened 2017-10-16
Choice rules

AntoniusW:
Enabling the use of choice rules in Alpha.


AntoniusW:
Codeclimate indicating failure, because a method requires more than 25 lines really seems a bit over the top.

lorenzleutgeb:
Nice! After looking at your latest changes regarding transformations, I also thought that choice rules were the next thing to approach.

lorenzleutgeb:
This branch needs a rebase. Should I do that or will you?

AntoniusW:
I will try to merge master and resolve the conflicts. Thanks for the offer, though.

AntoniusW:
@lorenzleutgeb I think the code is ready to merge, please have a look.

src/main/java/at/ac/tuwien/kr/alpha/grounder/atoms/HiddenAtom.java
```diff
@@ -0,0 +1,49 @@
...
+import at.ac.tuwien.kr.alpha.common.terms.Term;
+import at.ac.tuwien.kr.alpha.grounder.Substitution;
+
+import java.util.List;
+
+/**
+ * An internal ordinary atom, i.e., a BasicAtom that is internal and not reported in answer sets.
+ * Copyright (c) 2017, the Alpha Team.
+ */
+public class HiddenAtom extends BasicAtom {
```
lorenzleutgeb: I dislike the addition of `InternalAtom`:

 * Looking at the class, there is practically no behavior it adds to `BasicAtom` (calls to `super` everywhere).
 * We already have `BasicAtom.isInternal` which should "hide" an atom.

Proposed solution: Change the implementation of `BasicAtom.isInternal` so that it does not always return `false`, but some flag `private final boolean internal`. This flag should be `false` by default. Add a new constructor that allows to set it to `true` and use this instead of `HiddenAtom`.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ChoiceHead.java
```diff
@@ -0,0 +1,91 @@
...
+
+	public BinaryOperator getLowerOp() {
+		return lowerOp;
+	}
+
+	public BinaryOperator getUpperOp() {
+		return upperOp;
+	}
+
+	private final BinaryOperator lowerOp;
```
lorenzleutgeb: Please move all class member declarations to the top.

src/main/java/at/ac/tuwien/kr/alpha/grounder/atoms/HiddenAtom.java
AntoniusW: True, the `HiddenAtom` is like an ordinary atom that does not show up in answer sets.
By the way, why is the hiding not property of a `Predicate` but of an atom?

AntoniusW:
Regarding the second note, I think the current problem is that we actually want to say "rule head" but since this could so far only be an atom, we said `Atom` instead in all places. When realizing disjunction, the same problems will surface again: a disjunction of atoms may occur as a rule head but all those `Atom` methods cannot be used on it. Choices and (a limited form of) disjunction can be removed by translating the input program, thus it seems to me, that we could just introduce an empty interface `RuleHead` and both `Atom` and `ChoiceHead` simply extend/implement the `RuleHead` interface.

src/main/java/at/ac/tuwien/kr/alpha/grounder/atoms/HiddenAtom.java
lorenzleutgeb: I am not sure. But it would probably be better if we drop `Atom.isInternal` and instead replace `atom.isInternal()` by something like `isInternal(atom.getPredicate())` in `NaiveGrounder.assignmentToAnswerSet`, which then checks for some predefined instances of `Predicate`, just like in the `NaiveGrounder` constructor which calls `adaptWorkingMemoryForPredicate` for a similar set of predicates.

lorenzleutgeb:
I don't think that we would benefit from an empty interface. We'd probably end up with `instanceof` checks anyway. But we could introduce a `Head` type which then translates to a list of `Atom`s, where a non-disjunctive head translates to just a singleton list.

AntoniusW:
Done.

AntoniusW:
The AppVeyor build failed hitting some timeout and I cannot restart the build. @lorenzleutgeb could you restart it?

src/test/java/at/ac/tuwien/kr/alpha/solver/SolverTests.java
```diff
@@ -558,6 +567,58 @@ public void groundAtomInRule() throws IOException {
...
+
+	@Test
+	public void doubleChoiceRule() throws IOException {
+		Program program = parser.parse("{ a }. { a }.");
+		Solver solver = getInstance(new NaiveGrounder(program));
+		int countAnswerSets = 0;
+		while (solver.spliterator().tryAdvance(d -> { })) {
+			countAnswerSets++;
+		}
+		assertEquals(2, countAnswerSets);
```
lorenzleutgeb: There's no need to go so low-level here.

```java
assertEquals(2, solve("{ a }. { a }.").size());
```

src/main/java/at/ac/tuwien/kr/alpha/grounder/transformation/ChoiceHeadToNormal.java
```diff
@@ -0,0 +1,85 @@
...
+ * Copyright (c) 2017, the Alpha Team.
+ */
+public class ChoiceHeadToNormal implements ProgramTransformation {
+	public final static String PREDICATE_NEGATION_PREFIX = "_n";
+	private final NaiveGrounder grounder;
+
+	public ChoiceHeadToNormal(NaiveGrounder grounder) {
+		this.grounder = grounder;
+	}
+
```
lorenzleutgeb: Remove empty line.

src/main/java/at/ac/tuwien/kr/alpha/grounder/transformation/ChoiceHeadToNormal.java
```diff
@@ -0,0 +1,85 @@
...
+				ruleBody.addAll(choiceElement.conditionLiterals);
+
+				if (head.containsIntervalTerms()) {
+					throw new RuntimeException("Program contains a choice rule with interval terms in its head. This is not supported (yet).");
+				}
+
+				// Construct head atom for the guess.
+				Predicate headPredicate = head.getPredicate();
+				Predicate negPredicate = new Predicate(PREDICATE_NEGATION_PREFIX + headPredicate.getPredicateName(), headPredicate.getArity() + 1);
+				grounder.flagPredicateAsInternal(negPredicate);
```
lorenzleutgeb: I dislike this backdoor to the grounder, and also the possibly very large set of predicates that need to be checked for.

The grounder might as well check `predicate.getPredicateName().startsWith(PREDICATE_NEGATION_PREFIX)`.

Or we add `isInternal()` to `Predicate` ...

src/main/java/at/ac/tuwien/kr/alpha/common/Head.java
```diff
@@ -0,0 +1,101 @@
...
+
+import java.util.List;
+
+/**
+ * Represents the head of a rule, i.e., either a choice or a disjunction of atoms, but not both.
+ * For normal rules the disjunction contains only one atom.
+ *
+ * Copyright (c) 2017, the Alpha Team.
+ */
+public class Head {
```
lorenzleutgeb: What about making `Head` abstract and adding two non-abstract subclasses `DisjunctiveHead` and `ChoiceHead`? I dislike the pattern of having two properties of which one needs to be empty/null.

src/main/java/at/ac/tuwien/kr/alpha/grounder/NaiveGrounder.java
```diff
@@ -145,6 +148,14 @@ public NaiveGrounder(Program program, java.util.function.Predicate<Predicate> fi
 		}
 	}
 
+	private void applyProgramTransformations(Program program) {
+		// Transform choice rules.
+		new ChoiceHeadToNormal(this).transform(program);
+		// Transform intervals.
+		new IntervalTermToIntervalAtom().transform(program);
+	}
+
```
lorenzleutgeb: Remove empty line.

src/test/java/at/ac/tuwien/kr/alpha/solver/SolverTests.java
AntoniusW: I had my reasons for that as follows: choice rules introduce a rewriting with hidden predicates (basically `_na :- not a.  a :- not _na.` twice) and the number of answer-sets before projection of hidden predicates may be different than after projection.

The high-level collection of answer-sets is based on the projected answer-sets collected in a `Set<AnswerSet>`. So the solver might find two distinct answer-sets `{ _na }`and `{ _na' }` but after projection they both become `{ }`. The `Set<AnswerSet>` then swallows one of the answer-sets actually found be the solver. By goind low-level as in the test, it should be possible to observe them both, however.

This is the reason, why I think the high-level counting may not work as intended by the test. I'm happy to learn of a better method for checking the actually found answer-sets, though.

lorenzleutgeb: Oh, yeah. In this case you could use `solver.collectList()` which should give you a list containing two elements, both of which are equal to the empty answer set.

```
List<AnswerSet> expected = Arrays.asList(BasicAnswerSet.EMPTY, BasicAnswerSet.EMPTY);
List<AnswerSet> actual = solver.collectList();
assertEquals(expected, actual);
```

AntoniusW: The answer sets for this program indeed are `{}, { a }` but the test shall ensure that no further answer sets (which differ only on hidden atoms) are reported. So if there were two "same" answer sets, the `collectList()` should see them.

src/main/java/at/ac/tuwien/kr/alpha/grounder/transformation/ChoiceHeadToNormal.java
AntoniusW: Added `isInternal()` to `Predicate`, looks better now.

AntoniusW:
Addressed the raised issues, please check again.

PR #64, opened 2017-10-16
Introduce Evaluable interface for Predicates

lorenzleutgeb:
I reverted #61 and brought back the original changes to have more time for review.

 - [x] Remove warnings about unchecked cast(s).
 - [x] Research into possible heap pollution.
 - [x] Discuss open questions.

Blocked by #65


src/main/java/at/ac/tuwien/kr/alpha/common/predicates/Evaluable.java
```diff
@@ -0,0 +1,11 @@
+package at.ac.tuwien.kr.alpha.common.predicates;
+
+import at.ac.tuwien.kr.alpha.common.terms.Term;
+import at.ac.tuwien.kr.alpha.grounder.Substitution;
+
+import java.util.List;
+
+@FunctionalInterface
+public interface Evaluable {
```
AntoniusW: From what I understand this interface should cover atoms with a fixed interpretation similar as builtin atoms. Some other name might be more fitting then (maybe FixedInterpretationEvaluable?)

src/main/java/at/ac/tuwien/kr/alpha/common/predicates/ExternalNativePredicate.java
```diff
@@ -0,0 +1,36 @@
+package at.ac.tuwien.kr.alpha.common.predicates;
+
+import at.ac.tuwien.kr.alpha.common.terms.ConstantTerm;
+import at.ac.tuwien.kr.alpha.common.terms.Term;
+import at.ac.tuwien.kr.alpha.grounder.Substitution;
+
+import java.util.List;
+
+public class ExternalNativePredicate implements Predicate, Evaluable {
```
AntoniusW: What is native about this external predicate?

src/main/java/at/ac/tuwien/kr/alpha/common/predicates/TotalOrder.java
```diff
@@ -0,0 +1,85 @@
+package at.ac.tuwien.kr.alpha.common.predicates;
+
+import at.ac.tuwien.kr.alpha.common.terms.Term;
+import at.ac.tuwien.kr.alpha.grounder.Substitution;
+
+import java.util.List;
+
+public class TotalOrder implements Evaluable, Predicate {
```
AntoniusW: TotalOrder is more general than what this actually is (it is a certain instance of a total order). Since this boils down to the builtin atom, I suggest calling it similar to that.

src/main/java/at/ac/tuwien/kr/alpha/common/predicates/TotalOrder.java
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	public TotalOrder(String predicateName, boolean flip) {
+		if (!flip) {
+			this.predicateName = predicateName;
+			return;
+		}
+
+		switch (predicateName) {
+			case "=":
```
AntoniusW: I'd prefer the BINOP enum being used here, since this enum occurs at several positions (aggregates, choice rules) and the code seems to be better typed in the end than with strings.

src/main/java/at/ac/tuwien/kr/alpha/common/terms/FunctionTerm.java
```diff
@@ -116,23 +115,22 @@ public int hashCode() {
 
 	@Override
 	public int compareTo(Term o) {
+		if (o instanceof ConstantTerm) {
+			return 1;
+		}
 		if (!(o instanceof FunctionTerm)) {
-			throw new ClassCastException();
+			throw new UnsupportedOperationException("Can only compare function term to function term or constant term.");
```
AntoniusW: According to the ASP Core2 standard, all kinds of terms can be compared. Our previous orders did not take this into account, but we should use/support the official term ordering soonish.

src/main/java/at/ac/tuwien/kr/alpha/grounder/parser/ParseTreeVisitor.java
```diff
@@ -3,42 +3,92 @@
...
+	public ParseTreeVisitor(Map<String, Predicate> externals, boolean acceptVariables) {
+		this.externals = externals;
+		this.acceptVariables = acceptVariables;
+	}
+
 	private void notSupportedSyntax(RuleContext ctx) {
 		throw new UnsupportedOperationException("Unsupported syntax encountered: " + ctx.getText());
 	}
 
+	public Program translate(ASPCore2Parser.ProgramContext input) {
```
AntoniusW: What is translated here? ParseProgram might be a better name.

src/main/java/at/ac/tuwien/kr/alpha/grounder/parser/ParseTreeVisitor.java
```diff
@@ -3,42 +3,92 @@
...
+		return visitProgram(input);
+	}
+
+	public Set<AnswerSet> translate(ASPCore2Parser.Answer_setsContext input) {
```
AntoniusW: Similarly as above, ParseAnswerSet might be more descriptive.

src/test/java/at/ac/tuwien/kr/alpha/solver/SolverTests.java
```diff
@@ -27,174 +27,122 @@
...
-				.predicate("p").instance("1")
-				.build(),
-			new BasicAnswerSet.Builder(base)
-				.predicate("q").instance("3")
-				.predicate("p").instance("1").instance("2")
-				.build()
-		));
-
-		assertEquals(expected, solver.collectSet());
+			"q(X) :- dom(X), not p(X).",
```
AntoniusW: This is a really neat syntax now. Impressive!

src/main/java/at/ac/tuwien/kr/alpha/Alpha.java
```diff
@@ -0,0 +1,114 @@
...
+		}
+	}
+
+	public void register(Method method) {
+		this.predicateMethods.put(method.getName(), new ExternalEvaluable(method));
+	}
+
+	public void register(String name, java.util.function.Predicate<ConstantTerm> predicate) {
+		this.predicateMethods.put(name, new ExternalNativePredicate(name, predicate));
+	}
```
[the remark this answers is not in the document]
lorenzleutgeb: Yes. The `Predicate` lambda will be called whenever the truth of some atom over the predicate `"&" + name` is "queried". TBH I can't really think of any name that is much more "catchy". Do you have suggestions? Otherwise I'll just add some comment to clarify.

src/main/java/at/ac/tuwien/kr/alpha/Predicate.java
```diff
@@ -0,0 +1,4 @@
+package at.ac.tuwien.kr.alpha;
+
+public @interface Predicate {
```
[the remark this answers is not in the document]
lorenzleutgeb: This is a marker, to be used for discovery of external predicates via annotations. I should have added a comment, sorry.

In essence, it enables code like

```java
@at.ac.tuwien.kr.alpha.Predicate
public static boolean coolPerson(String nameOfPerson) { ... }
```

to be discovered. The interface is currently empty, because we're not using any options, it just marks methods to be registered for now.

src/main/java/at/ac/tuwien/kr/alpha/common/predicates/Evaluable.java
lorenzleutgeb: Yes, they have a fixed interpretation. Maybe only `FixedEvaluable` suffices? `FixedInterpretationEvaluable` is quite long :)

src/main/java/at/ac/tuwien/kr/alpha/common/predicates/ExternalNativePredicate.java
lorenzleutgeb: This class wraps `java.util.function.Predicate` which I coined "native" predicates ... I was in a hurry.

src/main/java/at/ac/tuwien/kr/alpha/common/predicates/TotalOrder.java
lorenzleutgeb: OK, so `BuiltinOrder`.

src/main/java/at/ac/tuwien/kr/alpha/common/predicates/TotalOrder.java
lorenzleutgeb: OK

src/test/java/at/ac/tuwien/kr/alpha/solver/SolverTests.java
lorenzleutgeb: Yeah. It's probably a bit slower, because the answer sets need to be parsed, but it is _much_ cleaner. I also thought about having our tests file based, for example as follows:

```
c :- not a, not b.

%< SNIP %< SNAP

{ a }
{ b }
{ c }
```

JUnit could then just try all of them.

src/main/java/at/ac/tuwien/kr/alpha/common/terms/FunctionTerm.java
lorenzleutgeb: Yeah, from what I see, the changes in this branch improve the situation, but we're not 100% there yet.

src/main/java/at/ac/tuwien/kr/alpha/Alpha.java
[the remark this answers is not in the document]
AntoniusW: I think it should be called `RegisterSomething` where `Something` is an appropriate name for what is registered here. At the moment, I am, however, really confused about the terminology. I tried to figure it out looking at classes in the package `alpha.common.predicate` but got even more confused. I think those class names should be cleaned up somehow. Currently, there is Evaluable, External, ExternalEvaluable (which does not implement the Evaluable interface?), and ExternalNativeEvaluable (which I guess is the implementation of an external in Java?). And then there is the `Method` from reflections used as a parameter above.

By the way, the `Evaluable` has a method `evaluate` and here you said the `Predicate` lambda will be used to decide the truth of an external atom. Should those two things not be the same or what is their difference?

Basically, what is needed for external atom evaluation is:
* an interface of one evaluation method that maps a list of k terms (input) to a list of l terms (output).
* an predicate class for external predicates which extends ordinary predicates with the information of how many input terms and how many output terms are there.
* a class storing the association of evaluation methods and external predicates (where evaluation methods are registered as above).

Maybe I am forgetting something now. I also assume those three types are already here, but maybe under a different name than I would expect and probably organized differently.

src/main/java/at/ac/tuwien/kr/alpha/grounder/parser/ParseTreeVisitor.java
AntoniusW: Ahh, right, my understanding at this point was wrong.
Btw, your explanation would make a good javadoc for this method. :-)

src/test/java/at/ac/tuwien/kr/alpha/solver/SolverTests.java
AntoniusW: Yes, one could put some tests in files. But I think it is more convenient to have them as source, because if a test fails, I can simply start debugging it with a right click. If it is file based, I assume there is only one entry point for all tests and this entry must be configured correctly to start the desired test. And currently I see no advantage of having each test in its own file(s).

lorenzleutgeb: Alright.

src/main/java/at/ac/tuwien/kr/alpha/Alpha.java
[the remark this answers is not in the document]
lorenzleutgeb: OK, then I would go for `registerPredicate`.

My picture of these things is as follows:

We have predicates, which are just words/names and a corresponding arity. They are represented by `Predicate`.

We distinguish between internal and external predicates:

 * For internal predicates, their interpretation (e.g. set of n-tuples of terms that make an atom over the predicate true) is defined through rules in the program.
 * For external predicates, their interpretation is computed by some interpretation function which is not part of the ASP program.

We further distinguish external predicates:

 * For external predicates with fixed interpretation the interpretation function takes n ground terms and returns a boolean truth value.
 * For external predicates with varying interpretation, the interpretation function takes n ground terms, and the current truth assignment and returns a boolean truth value.

Now, the change set that we are talking about here, introduces `FixedEvaluable` as an interface that external predicates with fixed interpretation should implement. Now we have some possibilities to do that:

 * Write a Java class that implements `Predicate` and `FixedEvaluable`. Encode its interpretation function in the `evaluate` method.
 * Just associate a name and an existing method, and generate an object of some type that implements `Predicate` and `FixedEvaluable`. The class `ExternalEvaluable` serves this purpose.  When it is being evaluated, then it "delegates" and calls the `Method` that was given at construction time. I might rename this class to something more descriptive based on our discussion, maybe `ExternalMethodEvaluable`.
 * Just associate a name an existing function. Since Java 8 allows for anonymous functions (lambda expressions) I thought that we might also support these. The principle is the same as for method references and the family of classes that extend `BaseExternalNativePredicate` serve this purpose. Java comes with `Predicate<T>`, `BiPredicate<T, U>`, `IntPredicate` and `LongPredicate` (I ignored `DoublePredicate` for now). So for all of them we have a corresponding class, that wraps an object of one of these types and creates a new object of a type that again implements `FixedEvaluable` and `Predicate`.

> By the way, the `Evaluable` has a method `evaluate` and here you said the `Predicate` lambda will be used to decide the truth of an external atom. Should those two things not be the same or what is their difference?

Whenever `FixedEvaluable.evaluate` is called, depending on which of the three above routes were taken to construct the predicate, the following might happen:

 * The object is of a type that has a "concrete" implementaion for `evaluate`, this will just be executed.
 * The object was constructed with a method reference, then the call to `evaluate` is delegated to that method.
 * The object was constructed with a function, then `evaluate` is delegated to the `test` method of the predicate.

So, up to now I only have worked on making "fixed interpretations" possible. I thought that having this makes sense on it's own, but also it is simpler than directly implementing atoms that bind new variables.

The interface `Evaluable` with its method that takes a list of terms and returns a list of terms, is there to be used for what you are sketching. However, we currently have no implementing classes. I might as well remove that interface.

In order to implement atoms that bind variables, I think I am missing some puzzle pieces. Up until now, the changes on a "technology" level kept me busy, and actually I wanted to get some working prototype as fast as possible. What I still do not understand is how the truth value of these external atoms is defined. Suppose i have `employee(Y) :- employee(X), managerOf[Y](X)` (if X is an employee, and Y is the manager of X, then Y is also an employee), then how do I determine whether `manager[Y](X)` is true or false? Also, must these external atoms always map to one single term? What if `X`has more than one manager?
Probably this is all covered by the definition of external atoms. Could you refer me to some paper?

What still confuses me is why these constructions are called "external atom" and not "external predicate". This might however indicate a fundamental flaw in my thinking. For me, predicates that use external functions to define their interpretations can very well be called external predicates. However, only atoms may bind variables, so we then also get atoms that use functions to bind variables, which would be something else than an external predicate, namely an external atom.

Also, how does all this relate to function terms? Wouldn't an external atom that binds exactly one term have the same power as a function?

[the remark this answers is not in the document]
AntoniusW: Ok now there are several separate poins to discuss. Let me enumerate those issues for better reference:

1. Method name:
I think `registerPredicate` is an improvement over the current name, but if I understand correctly, then this method also registers the implementation of the predicate, so I would suggest `registerPredicateImplementation`.

2. Predicate notion:
A predicate is just a name and an arity, we agree on that. This is my reason why the method under 1. should not just be `registerPredicate` because it does more than just saying there exists a name with such arity.

3. Distinction internal and external predicates:
I basically agree but would like to add that internal predicates are usually defined by the rules of the ASP program but also may be builtin and have a fixed interpretation (like the `=` predicate, for example).

4. Should external atoms receive the current interpretation:
This is an important question and I thought that we decided on No. If we decide that external atoms may depend on the current interpretation, then we have a system similar to HEX, which is very powerful but also harder to evaluate. If the truth of the external atom depends on an interpretation, we have to re-evaluate the external atom every time the interpretation changes (conceptually, in practice some evaluations can be avoided, but in the worst case we have to). This then implies that evaluating external atoms cannot be done simply during the grounding but much more work is needed. In fact, something similar as in the `alphahex` branch has to happen where Alpha is paired with the HEX solver and quite some necessary work is done in HEX (like computing a full grounding of certain parts of the input program).
Also note that terms have a fixed interpretation (since ASP is based on Herbrand interpretations where each term is interpreted by itself). So for an external atom `&external( term1, f(term2) )` the truth of the whole atom may depend on the implementation of the external atom, but its input will always be the same, namely ( term1, f(term2) ) regardless of the current interpretation.
Summarizing, I strongly suggest that we only allow `FixedEvaluables`.

5. Output of external atoms:
I think an external atom can give us some output (maybe even a set of outputs) in the form or terms. So, an external atom is a mapping from `k` terms to a set of `l`terms. From the theoretical view this is simply a `k+l`-ary Boolean function. Evaluating an external atom with `k` input terms then results in a tuple (or set of tuples) of `l` ouput terms where the reading is that the `k+l`-ary Boolean function is true for (each combination of) the input terms and the output terms and false for all other combinations of the given input terms with any possible output terms.

6. Multiple ways to create the implementation of external atoms:
I agree that for usability it is good to have multiple ways to actually implement the external. At the moment, I do not understand how the anonymous functions should work. I think it should always be a function from `List<Term>` to `List<Term>` (or rather to `Set<List<Term>>`, see below). How come `BiPredicate`, `IntPredicate` and so on into play here?

7. The missing piece on the `Evaluable` interface:
A function `f` mapping `k` input terms to `l` output terms can be represented as a Boolean function `g` on `k+l` input terms as follows:
`f(in_1, ..., in_k) = (out_1, ..., out_l)` iff `g(in_1, ..., in_k, out_1, ..., out_l)` is true.
So, conceptually we use external atoms in the style of `g` but to evaluate them we use the implementation of `f`.

8. Variable binding:
I think this should be done purely by the grounder and not inside the external atoms. So the grounder will ensure that an external atom is called only after all its input terms are bound and take the return values of the external atom and bind variables at output positions accordingly.

9. On the concrete example:
`employee(Y) :- employee(X), managerOf[Y](X)`
First, we have to clear what is input and output. In HEX the part between `[ ]` is input and between `( )` is ouput -- I would stick to that. So, let me correct the rule to:
`employee(Y) :- employee(X), &managerOf[X](Y).`
Now, the grounder looks for bindings of `X` and for each binding of `X` it will evaluate the implementation of `&managerOf` and get in return a set of bindings for `Y`. So the type of our implementation should be `Set<List<Term>> method(List<Term>)`. So for each binding `X=a` we call `method(a)` and get a set `{ (b), (c), (d) }` of all managers of `a`; in this case `b`, `c` and `d` are all managers of `a`. Now the grounder can bind `Y`and it generates a new ground instance of this rule for every binding (so we get 3 ground instances in this example).

10. References:
* http://www.kr.tuwien.ac.at/staff/redl/publications/ekrsw2017-rw.pdf
This is a summer school article introducing HEX and its foundations. It also contains some small examples but does not go into the details of the implementation.
* https://www.ijcai.org/proceedings/2017/0141.pdf
This shows how one can combine HEX and lazy grounding using Alpha (where the truth value of external atoms depends also on the current assignment). Although the Algorithm is not far from Alpha, there are several notions where no algorithm is given but which must be computed. These notions are input-safe domain, external input assignment, external evaluation function, and on top of that, all implementations of external atoms must be assignment-monotonic. Hence, I strongly suggest to not let external atoms depend on the current assignment but only on the given term input.

11. I think we could use a concrete example for discussion and later for testing. What did you have in mind when starting with all of this?

12. Why it is called external atom and not external predicate:
Well, the predicate is just a name and some arity. In order to speak about truth, we need some terms (of correct arities), hence we get an atom.

13. Connection to function terms:
In general, an interpretation associates each function term with one element from the universe. So yes, what the external atom is doing could be seen as a just a certain interpretation of a function term. But in logic programming and ASP we use Herbrand interpretations and those have a fixed interpretation for constants and function terms. In an Herbrand interpretation the universe always is the set of all terms and each constant `c` is mapped to the element `c` of the universe. Likewise, every function term `f(t_1,...,t_n)` is mapped to the element `f(t_1,...,t_n)` of the universe. Jaques Herbrand could show that for every first order formula `F` it holds that `F` is satisfiable if-and-only-if there exists a Herbrand interpretation that satisfies `F`. In essence, this allows us to decide validity of a formula by only looking at its syntactic form. Now, one might be able to realize external atoms by allowing for certain function symbols to not be interpreted in Herbrand-style, but then I would expect that things get ugly.

14. I think this work you do is totally worth it and it is great to see that development happen!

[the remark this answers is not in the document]
lorenzleutgeb: Thank your the thorough answer! I think we agree on most of the items, especially regarding (4.) I also do not want to go down the rabbit hole of passing around the current assignment/interpretation.

#### Regarding (6.):
> How come `BiPredicate`, `IntPredicate` and so on into play here?

Take for example the rule `reachable(A,C) :- &connected(A,B), &connected(B,C).` and further `:-
 connected(A,B), not node(A), not node(B)`, and assume that we have all objects that are nodes introduced to the program via facts. However, the question of answering whether two of these nodes are connected should be answered by a Java lambda expression of type `BiPredicate<Node, Node>`, that takes the two nodes, fiddles around with them in unexpected ways, and then gives us the answer whether two nodes (substituted for `A`, `B` and `C`) are connected. This comes very close to what you suggested in (5.), namely

> From the theoretical view this is simply a k+l-ary Boolean function.

So, here you have that function. In this special case k = l = 1. The difference to functions like you sketched in (9.), namely

> `Set<List<Term>> method(List<Term>)`

is that our `BiPredicate<Node,Node>` cannot return terms, so we cannot do variable binding as you suggest in (8.). I expect that these types of "interpretation functions" are easier to understand and implement, as they leave the assigning of variables/the binding for the grounder. Still, I think that even a simple use case as the one above is perfectly fine and makes sense to me.

This is specifically the reason why I implemented them first. I did not want to bother handling returned terms and binding them for a given rule.

#### Regarding (9.)
What would `Set<List<Term>> method(List<Term>)` return if there is no manager for some binding for `X`? The empty set? This way it should be relatively straight forward to implement.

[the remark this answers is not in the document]
AntoniusW: Thanks for the explanation regarding 6.  Allowing these as special cases seems okay.

Regarding 9. yes, exactly it would be the empty set.
In case of an external atom with no return value it would be a singleton set containing an empty list of terms, i.e., the term `()`, if the atom is true, and the empty set if the atom is false.

src/main/java/at/ac/tuwien/kr/alpha/common/Interner.java
```diff
@@ -0,0 +1,26 @@
+package at.ac.tuwien.kr.alpha.common;
+
+import java.lang.ref.WeakReference;
+import java.util.WeakHashMap;
+
+public class Interner<T> {
+	private WeakHashMap<T, WeakReference<T>> pool = new WeakHashMap<>();
+
+	public synchronized T intern(T object) {
+		T res;
```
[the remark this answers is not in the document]
lorenzleutgeb: * [Java Object Interning](https://www.javacodegeeks.com/2014/03/java-object-interning.html) (this is where the code comes from)
 * [Understanding Weak References](https://community.oracle.com/blogs/enicholas/2006/05/04/understanding-weak-references)

Interners allow for reference equality checks, and (maybe) save some memory.

In essence, we are using weak references here, because there is no point in keeping objects interned when there are no other references to it. If the objects we are referring to are not (strongly) referred by anything else, then they might just "disappear" from the interner.

Conversely, if we would not use weak references, then we would need additional logic that decides when an interned object should be removed from the interner, e.g. we would have to track when for example an instance of `ConstantTerm` can be garbage collected. We do not want that.

lorenzleutgeb:
I added `ExternalAtom`. Still some failing tests in `AlphaTest`. My understanding of the grounder is not solid enough to find the culprit right now, I will reconsider it.

lorenzleutgeb:
@AntoniusW, please give it another look.

src/main/java/at/ac/tuwien/kr/alpha/common/predicates/BaseExternalNativePredicate.java
```diff
@@ -0,0 +1,56 @@
...
+					"Expected only constants as input for " + getPredicateName() + ", but got " +
+						"something else at position " + i + "."
+				);
+			}
+
+			constants.add((ConstantTerm) terms.get(i));
+		}
+
+		try {
+			return test(constants) ? TRUE : FALSE;
```
AntoniusW: Why does `test` not return a `Set<List<ConstantTerm>>`?

src/main/java/at/ac/tuwien/kr/alpha/common/predicates/Evaluable.java
```diff
@@ -0,0 +1,16 @@
...
+
+import at.ac.tuwien.kr.alpha.common.terms.ConstantTerm;
+import at.ac.tuwien.kr.alpha.common.terms.Term;
+
+import java.util.List;
+import java.util.Set;
+
+import static java.util.Collections.*;
+
+public interface Evaluable extends Predicate {
```
AntoniusW: How about renaming `Evaluable` to `InterpretationAgnosticPredicate` (or `FixedValuePredicate`)?

src/main/java/at/ac/tuwien/kr/alpha/common/predicates/ExternalFixedEvaluable.java
```diff
@@ -0,0 +1,81 @@
...
+import at.ac.tuwien.kr.alpha.common.terms.ConstantTerm;
+import at.ac.tuwien.kr.alpha.common.terms.Term;
+import org.apache.commons.lang3.ClassUtils;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.List;
+import java.util.Set;
+
+public class ExternalFixedEvaluable implements Predicate, Evaluable {
```
AntoniusW: Maybe call this `ExternalPredicate` ?

src/main/java/at/ac/tuwien/kr/alpha/common/predicates/ExternalNativeBiPredicate.java
```diff
@@ -0,0 +1,23 @@
+package at.ac.tuwien.kr.alpha.common.predicates;
+
+import at.ac.tuwien.kr.alpha.common.terms.ConstantTerm;
+
+import java.util.List;
+
+public class ExternalNativeBiPredicate<T, U> extends BaseExternalNativePredicate {
```
AntoniusW: Do we have non-`Native` predicates? I suggest simply dropping `Native` in the name.

lorenzleutgeb:
New restructuring for terms:

predicates

src/main/java/at/ac/tuwien/kr/alpha/common/predicates/ExternalNativeBiPredicate.java
lorenzleutgeb: Done.

src/main/java/at/ac/tuwien/kr/alpha/common/predicates/ExternalFixedEvaluable.java
lorenzleutgeb: This would clash with `ExternalNativePredicate`, now called `ExternalPredicate` (according to your suggestion). So I went for `ExternalMethodPredicate`.

src/main/java/at/ac/tuwien/kr/alpha/common/predicates/Evaluable.java
lorenzleutgeb: Renamed it to `FixedInterpretationPredicate`, hope that is fine.

src/test/java/at/ac/tuwien/kr/alpha/AlphaTest.java
```diff
@@ -0,0 +1,200 @@
...
+		return term == 1;
+	}
+
+	@at.ac.tuwien.kr.alpha.Predicate
+	public static boolean isFoo(Integer a) {
+		return a == 0xF00;
+	}
+
+	@at.ac.tuwien.kr.alpha.Predicate
+	public static boolean thinger(Thingy thingy) {
```
AntoniusW: I believe I just now understand what this enables us to do. Maybe that's worth a paper!

src/test/java/at/ac/tuwien/kr/alpha/AnswerSetsParser.java
```diff
@@ -0,0 +1,95 @@
...
+	public static Set<AnswerSet> parse(CharStream stream) {
+		try {
+			return SINGLETON.parseDynamic(stream);
+		} catch (IOException e) {
+			throw new RuntimeException(e);
+			//return null;
+		}
+	}
+
+	public Set<AnswerSet> parseDynamic(CharStream stream) throws IOException {
```
AntoniusW: Is there a reason why this cannot be a static method?

src/test/java/at/ac/tuwien/kr/alpha/solver/SolverTests.java
```diff
@@ -27,174 +27,122 @@
...
+import static java.util.Collections.emptySet;
+import static java.util.Collections.singleton;
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
 
 public class SolverTests extends AbstractSolverTests {
-	/**
-	 * Sets the logging level to TRACE. Useful for debugging; call at beginning of test case.
-	 */
-	private static void enableTracing() {
```
AntoniusW: Although `enableTracing` and `enableDebugLog` are not used, they sometimes are handy to have when a test fails. I would like them to remain (somewhere).

src/main/java/at/ac/tuwien/kr/alpha/common/predicates/BaseExternalPredicate.java
```diff
@@ -0,0 +1,42 @@
...
+					"Expected only constants as input for " + getPredicateName() + ", but got " +
+						"something else at position " + i + "."
+				);
+			}
+
+			constants.add((ConstantTerm) terms.get(i));
+		}
+
+		try {
+			return test(constants) ? TRUE : FALSE;
```
AntoniusW: This implies that none of the externals derived from `BaseExternalPredicate` can bind variables. Fine for me.

lorenzleutgeb: Yes. All classes from the `Predicate` family just return a `boolean`, this should be fine.

AntoniusW: In the light of this I better understand the class hierarchy. I was wondering why `ExternalBindingMethodPredicate` was not a subtype of `BaseExternalPredicate`. Maybe a more intuitive naming would be to rename `BaseExternalPredicate` to `ExternalNoOutputPredicate` (or `ExternalNoOutputBasePredicate`) and `ExternalBindingMethodPredicate` to `ExternalOutputPredicate`?

src/main/java/at/ac/tuwien/kr/alpha/grounder/NaiveGrounder.java
```diff
@@ -290,24 +292,26 @@ public float computeAverageNoGoodLength() {
...
+
 		for (NonGroundRule nonGroundRule : rulesFromProgram) {
 			// Generate nogoods for all rules that are ground already.
 			if (!nonGroundRule.isOriginallyGround()) {
 				continue;
 			}
+
 			// Check if rule contains intervals (but is otherwise ground).
-			if (nonGroundRule.containsIntervals()) {
+			if (nonGroundRule.containsIntervals() || nonGroundRule.containsExternals()) {
```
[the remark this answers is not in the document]
lorenzleutgeb: this might be a misunderstanding on my side, however: Just as an interval expands to multiple bindings/substitutions (for example `a(1..3)` will expand to `a(X)` for `X = 1`, `X = 2`, `X = 3`), an external atom might also expand to multiple bindings/substitutions (for example `&coolNumbers[X](3)` might result in the same three possible bindings for `X`). At least that#s how I understand it. So therefore, we do the same for intervals and externals: We "unwrap" them and get multiple substitutions.

src/main/java/at/ac/tuwien/kr/alpha/grounder/NaiveGrounder.java
```diff
@@ -403,19 +412,26 @@ private void register(Iterable<NoGood> noGoods, Map<Integer, NoGood> difference)
...
 		ArrayList<Integer> bodyAtomsPositive = new ArrayList<>();
 		ArrayList<Integer> bodyAtomsNegative = new ArrayList<>();
 		for (Atom atom : nonGroundRule.getBodyAtomsPositive()) {
-			if (atom instanceof BuiltinAtom) {
+
+			if (atom instanceof ExternalAtom) {
+				ExternalAtom external = (ExternalAtom) atom;
+
+				if (external.hasOutput()) {
+					continue;
```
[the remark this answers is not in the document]
lorenzleutgeb: Externals with output just bind their output terms, so they already are covered when we have obtained a ground substitution. Conversely, we will never reach a ground substitution, if an external atom has output terms but does not provide any bindings for them.

External atoms that have no output terms have no say in the process of finding a substitution, they only read the arguments they are given and evaluate them, they will never bind anything. So these need to be evaluated for every ground substitution we have obtained.

src/test/java/at/ac/tuwien/kr/alpha/AlphaTest.java
lorenzleutgeb: Do you have any publication outlets (conferences etc.) in mind that would be interested in these kind of things? I'd like to get a feel for what is published there.

src/test/java/at/ac/tuwien/kr/alpha/solver/SolverTests.java
lorenzleutgeb: OK, brought them back.

src/main/java/at/ac/tuwien/kr/alpha/Alpha.java
```diff
@@ -0,0 +1,157 @@
...
+
+	public Alpha(String grounderName) {
+		this(grounderName, "default");
+	}
+
+	public Alpha() {
+		this("naive");
+	}
+
```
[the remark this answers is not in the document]
lorenzleutgeb: Yes, this will eventually evolve into the outwards facing API, but I expect more changes to follow, so I would prefer to get this branch merged before, and iterate on it afterwards. We can always refine it.

I already have a handful of other changes that depend on this PR, including code that better showcases how such an "outwards facing" API might be used.

src/main/java/at/ac/tuwien/kr/alpha/common/Symbol.java
```diff
@@ -56,6 +49,17 @@ public int hashCode() {
...
 	@Override
 	public int compareTo(Symbol o) {
-		return ComparisonChain.start().compare(symbol, o.symbol).compare(arity, o.arity).result();
+		int result = symbol.compareTo(o.symbol);
+
+		if (result != 0) {
+			return result;
+		}
+
+		return arity - o.arity;
```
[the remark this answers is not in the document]
lorenzleutgeb: Done.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalAtom.java
```diff
@@ -0,0 +1,131 @@
...
+	@Override
+	public Atom substitute(Substitution substitution) {
+		return new ExternalAtom(predicate, input.stream().map(t -> {
+			return t.substitute(substitution);
+		}).collect(Collectors.toList()), output, negated);
+	}
+
+	@Override
+	public int compareTo(Atom o) {
+		// TODO: Implement
```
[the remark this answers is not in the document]
lorenzleutgeb: I moved `compareTo` up to the `Atom` interface, this resolves the issue.

src/main/java/at/ac/tuwien/kr/alpha/grounder/NaiveGrounder.java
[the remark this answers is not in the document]
AntoniusW: Ah, now I see. This place here is treating a special case, namely if the rule contains no ordinary atom with a variable that starts the grounding/substitution-finding of the rule. So I guess the check should not just be for the presence of an output-variable in the atom, but for the non-presence of any ordinary variable in an ordinary atom.

Maybe the code currently does no harm when called for all rules that contain external atoms. But I am not fully sure.

src/main/java/at/ac/tuwien/kr/alpha/common/predicates/BaseExternalPredicate.java
lorenzleutgeb: Is there any particular reason why you suggest to go from "binding" to "output"? I think with this we might introduce even more words that might be juggled around, and that "binding" captures quite well what it outputs.

In the scheme I proposed (last), the class hierarchy branches so that every instantiable class with a name matching `ExternalBinding.*Predicate` will call some external method that will return a set of lists of terms, and everything that just has `External.*Predicate` will call some external method that returns a boolean, which will then be converted into a set of lists of terms.

So the reason why `ExternalBindingMethodPredicate` is not a subtype of `BaseExternalPredicate` is precisely that it will return a set of lists of terms (e.g. bindings), while the others, that are below `BaseExternalPredicate` will not (they just return a boolean).

src/test/java/at/ac/tuwien/kr/alpha/AlphaTest.java
AntoniusW: There might be some, it depends on the actual use cases we have. We could discuss that directly by email.

src/main/java/at/ac/tuwien/kr/alpha/grounder/NaiveGrounder.java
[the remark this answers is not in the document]
AntoniusW: If the external evaluates to false, then no rule with this external can fire, so we may directly stop finding a full substitution since the rule(s) will be skipped anyways. Returning `emptyList()` does the stopping.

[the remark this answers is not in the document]
lorenzleutgeb: This is done here: https://github.com/alpha-asp/Alpha/pull/64/files#diff-3bb08e391a45e7c42bba1e6471511498R586 (with some horrible variable naming)

src/main/java/at/ac/tuwien/kr/alpha/grounder/NaiveGrounder.java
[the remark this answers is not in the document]
lorenzleutgeb: OK. We should test this. If you can construct a test case that shows incorrectness of this implementation from the top of your head, this would be very helpful. I will try to construct one otherwise.

Probably then I need to just refine/reformulate the check to what you suggested.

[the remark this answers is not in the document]
lorenzleutgeb: OK, I just had another look, and I don't see how this might lead to problems ... Could you provide a "counterexample", e.g. showcase how this behaves wrongly?

src/main/java/at/ac/tuwien/kr/alpha/Alpha.java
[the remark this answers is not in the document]
AntoniusW: Ok sounds reasonable to refine it later when things have settled. I also have some changes looking forward to this major PR to come into effect.

src/main/java/at/ac/tuwien/kr/alpha/common/predicates/BaseExternalPredicate.java
AntoniusW: There is some reason, namely that I need to make the grounder care better for what are binding variables and in what order the grounding of a rule body should proceed. I'm thinking of each atom reporting which variables it can bind and which others must be bound already. In case of ordinary atoms this is simple, all positive atoms bind variables, negative ones never bind. But in case of arithmetics and aggregates this is more involved and I'd like to try a better solution than the current one (which is even not according to the standard). Maybe one which allows to rewire the evaluation of atoms depending on things like precision and quantity.

So, "binding" for me has to do with variable binding (located purely in `NaiveGrounder`), which is actually nothing that I expected the external atoms to do. You wrote code, which does variable binding (`getSubstitutions`), so it currently is not so separated as I had in mind and I would like to keep your code.
Would renaming `BaseExternalPredicate` to `ExternalNonBindingBasePredicate` be acceptable for you? If not, then we can also keep it as is.

src/main/java/at/ac/tuwien/kr/alpha/grounder/NaiveGrounder.java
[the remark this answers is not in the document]
AntoniusW: I think it will work with the current implementation of the grounding, but might fail if we change that later. In any case it is unnecessary to do for all rules. A `// FIXME: should only be done for rules with externals where the rule contains no variable bound by an ordinary atom.` should be enough for now.

src/main/java/at/ac/tuwien/kr/alpha/grounder/NaiveGrounder.java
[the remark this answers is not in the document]
AntoniusW: Oh, I was not aware of that, sorry.

src/main/java/at/ac/tuwien/kr/alpha/grounder/parser/ParseTreeVisitor.java
```diff
@@ -37,8 +37,8 @@ public ParseTreeVisitor(Map<String, FixedInterpretationPredicate> externals, boo
 		this.acceptVariables = acceptVariables;
 	}
 
-	private void notSupportedSyntax(RuleContext ctx) {
-		throw new UnsupportedOperationException("Unsupported syntax encountered: " + ctx.getText());
+	private UnsupportedOperationException notSupported(RuleContext ctx) {
+		return new UnsupportedOperationException("Unsupported syntax encountered: " + ctx.getText());
```
AntoniusW: 👍

AntoniusW:
@lorenzleutgeb Can you look into the failing test cases?

lorenzleutgeb:
OK, I think I resolved the ugly hack in the special case that we discussed earlier. Along the way I encountered some strange behavior when there are ground rules, and when I attempted to lift the restriction that builtin atoms cannot create bindings. I will probably showcase these two once this PR is settled.

PR #65, opened 2017-10-16
Create CONTRIBUTING.md

lorenzleutgeb:
Since we had troubles managing our PRs in the past, mainly because there was no defined process IMO, I propose the addition of contributing guidelines to the project.

I will suspend any work on the project until we have agreed on contributing guidelines.


AntoniusW:
This is a good observation and I think the guidelines are fine. We can change them if they turn out to cause problems, but for now, I think, the described process is very good.

PR #67, opened 2017-10-26
Fix some smells

lorenzleutgeb:
These are some changes that accumulated over the last few weeks. They do not fix particular bugs, but improve readability/maintainability/performance.


src/main/java/at/ac/tuwien/kr/alpha/grounder/parser/ProgramParser.java
```diff
@@ -24,6 +24,22 @@ public ProgramParser() {
...
+	public Program parse(String s) {
+		try {
+			return parse(CharStreams.fromString(s));
+		} catch (IOException e) {
+			// In this case we assume that something went fundamentally
+			// wrong when using a String as input. The caller probably
+			// assumes that I/O on a String should always be fine.
+			throw new RuntimeException("Encountered I/O-related exception while parsing a String.", e);
+		} catch (RecognitionException | ParseCancellationException e) {
+			// If there were issues parsing the given string, we
```
AntoniusW: At least one test case (`ParserTest.parseBadSyntax`) is expecting the RecognitionException to be returned. It is not fully clear to me, why hiding the parser exceptions behind an `IllegalArgumentException` is good? I would have expected that more specific exceptions are preferable.

lorenzleutgeb: My idea was that in case the parser cannot parse `s` it is an illegal argument. Yes, this is a more general exception. However, we supply a more helpful message ("Could not parse answer sets.") on top of the more concrete `RecognitionException`. Further we pass `e` as a cause, so traces will be `IllegalArgumentException ... Could not parse answer sets ... caused by RecoginitionException` which does not completely hide the underlying cause. This is called ["Chained Exceptions"](https://docs.oracle.com/javase/tutorial/essential/exceptions/chained.html) and actually a common practice.

AntoniusW: Good point, in this case the test case should be adapted to fit the new exception(s).

lorenzleutgeb: Done

PR #68, opened 2017-10-26
Fix external atom variable bindings

### LINKED ISSUE #66: Fix external atom variable bindings.
External atoms have two open issues.
1. The binding of variables they report seems to contain a bug. External atoms have their input always non-binding, since they cannot be queried without some concrete input. I think for external atoms it should be as follows:
- If the external atom is negative, then all variables of input and output are non-binding and there are no binding variables (like for ordinary atoms).
- If the external atom is positive, then variables of input are non-binding and variables of output are binding.

2. In the `NaiveGrounder` line 585 there is a check
`if (substitutions.isEmpty() && rule.isBodyAtomPositive(atomPos))`
where it probably should just check for `isEmpty()`. The current check seems to solve an underlying bug where `ExternalAtom.getSubstitutions` should, in case of no bound variables, still return a singleton list containing only the partialSubstitution it received as input. It currently does not, because probably some subclass of FixedInterpretationPredicate.evaluate returns an empty set (of lists of constants) for when it finds no bindings, when instead it should return a singleton set containing an empty list of constants.
So an external that is true but has no output should yield a singleton set containing an empty list (of constants) while an external that is false (and hence has not output) should yield an empty set (of lists of constants).

lorenzleutgeb:
Fixes #66

### DIFF
--- a/src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalAtom.java
+++ b/src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalAtom.java
@@ -12,6 +12,7 @@ import java.util.*;
 import java.util.stream.Collectors;
 
 import static java.util.Collections.emptyList;
+import static java.util.Collections.unmodifiableList;
 
 public class ExternalAtom implements Literal {
 	private final List<Term> input;
@@ -91,29 +92,26 @@ public class ExternalAtom implements Literal {
 
 	@Override
 	public List<VariableTerm> getBindingVariables() {
-		if (negated) {
-			// Negative literal has no binding variables.
-			return Collections.emptyList();
-		}
-		LinkedList<VariableTerm> bindingVariables = new LinkedList<>();
-		for (Term term : input) {
-			bindingVariables.addAll(term.getOccurringVariables());
-		}
-		bindingVariables.addAll(output);
-		return bindingVariables;
+		// If the external atom is negative, then all variables of input and output are non-binding
+		// and there are no binding variables (like for ordinary atoms).
+		// If the external atom is positive, then variables of output are binding.
+		return negated ? emptyList() : unmodifiableList(output);
 	}
 
 	@Override
 	public List<VariableTerm> getNonBindingVariables() {
-		if (!negated) {
-			// Positive literal has only binding variables.
-			return Collections.emptyList();
-		}
+		// External atoms have their input always non-binding, since they cannot
+		// be queried without some concrete input.
 		LinkedList<VariableTerm> nonbindingVariables = new LinkedList<>();
 		for (Term term : input) {
 			nonbindingVariables.addAll(term.getOccurringVariables());
 		}
-		nonbindingVariables.addAll(output);
+
+		// If the external atom is negative, then all variables of input and output are non-binding.
+		if (negated) {
+			nonbindingVariables.addAll(output);
+		}
+
 		return nonbindingVariables;
 	}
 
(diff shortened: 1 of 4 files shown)


src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalAtom.java
```diff
@@ -91,29 +92,26 @@ public boolean isInternal() {
...
-		LinkedList<VariableTerm> bindingVariables = new LinkedList<>();
-		for (Term term : input) {
-			bindingVariables.addAll(term.getOccurringVariables());
-		}
-		bindingVariables.addAll(output);
-		return bindingVariables;
+		// If the external atom is negative, then all variables of input and output are non-binding
+		// and there are no binding variables (like for ordinary atoms).
+		// If the external atom is positive, then variables of output are binding.
+		return negated ? emptyList() : unmodifiableList(output);
```
AntoniusW: I just now realized that `output` is a `List<VariableTerm>` and I guess this will make problems if one wants to use an external atom whose output is partially grounded.
For example as in the rule
`outgoing13(X) :- &getLargeGraphEdges[](13,X).`
which would be an external atom without input and two outputs where we are currently just interested in outgoing edges of node `13`. Does this work with the current `ExternalAtom`? I assume it does not and the fix probably would require that `output` is a list of `Terms`. Here you then have to scan that list for variables and return only those in case the atom is not negated.

This has also implications for the code where the binding occurs, because not every output is a variable and also not all returned tuples match with the grounded part. Those two conditions must be taken into account there (this probably occurs in several subclasses).

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalAtom.java
```diff
@@ -91,29 +92,26 @@ public boolean isInternal() {
...
+		// be queried without some concrete input.
 		LinkedList<VariableTerm> nonbindingVariables = new LinkedList<>();
 		for (Term term : input) {
 			nonbindingVariables.addAll(term.getOccurringVariables());
 		}
-		nonbindingVariables.addAll(output);
+
+		// If the external atom is negative, then all variables of input and output are non-binding.
+		if (negated) {
+			nonbindingVariables.addAll(output);
```
AntoniusW: Similar as above, the output probably must be a list of `Terms` and scanned for occurring variables.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalAtom.java
lorenzleutgeb: Turns out that I confused input and output in the grammar. I just pushed abe82e7 to interpret external atoms as `&p[I1, I2, ..., In](O1, O2, ..., On)` where `I` are input and `O` are output.

Indeed, I think that we should forbid to have anything else than *variables* in the output list (for now) for the sake of simplicity.

I would interpret `&p[a](b,X)` as a type of filter. E.g. we run the function behind `&p` with input `a` and then "unify" it with `b, X` effectively discarding all output that does not have `b` at position 1. Equivalently: `&p[a](Y,X), Y = b`.

We can add this in a back-compatible manner any time.

AntoniusW: Forbidding anything except variables is okay for now and I think that later we should have the code for filtering directly in the `NaiveGrounder`, so that all code dealing with variables and substitutions is in one place. This probably means that the current code dealing with substitutions at the external atoms will move, too.

I agree with your interpretation of bound variables as some kind of filter. Actually, this gives rise to a simple rewriting for allowing terms as output, just the same way as in your example.

PR #73, opened 2017-10-29
Restructure AbstractSolverTests

lorenzleutgeb:
Do not test so many configurations by default, just one. Allow setting configurations to be tested via properties.

This also allows for testing multiple grounder implementations.


src/test/java/at/ac/tuwien/kr/alpha/solver/PigeonHoleTest.java
```diff
@@ -142,7 +118,7 @@ public void test55Pigeons60Holes() throws IOException {
 	}
 
 	/**
-	 * Tries to solve the problem of assigning P pigeons to H holes.
+	 * Tries to collectSet the problem of assigning P pigeons to H holes.
```
AntoniusW: Revert please. :-)

Issue #72, opened 2017-10-29
Lift restrictions on external atoms.


AntoniusW:
Currently, external atoms have the following limitations:

1. An external must have some input and an external like `&getFromExtern[](X)` is not possible.
2. An rule comprised of only an external whose input is are constants only will yield wrong answer sets (it is never evaluated). A rule like `p(X) :- &getFromExtern[0](X).` will not be evaluated correctly.
3. An external cannot have non-variable output like in `&getFromExtern[0](3)`.

The following code for testing may be easily adapted to showcase each of the restrictions.

	public static Set<List<ConstantTerm<Integer>>> returnTestNodes(int drop) {
		Set<List<ConstantTerm<Integer>>> ret = new LinkedHashSet<>();
		for (int i = 0; i < 5; i++) {
			List<ConstantTerm<Integer>> termList = new LinkedList<>();
			termList.add(ConstantTerm.getInstance(i + 3));
			ret.add(termList);
		}
		return ret;
	}

	@Test
	public void externalAtomOutput() throws Exception {
		Alpha system = new Alpha();
		system.registerBinding(this.getClass().getMethod("returnTestNodes", int.class));

		Set<AnswerSet> actual = system.solve("p(0). node(X) :- p(Y), &returnTestNodes[Y](X).").collect(Collectors.toSet());
		Set<AnswerSet> expected = AnswerSetsParser.parse("{ p(0), node(3), node(4), node(5), node(6), node(7) }");
		assertEquals(expected, actual);
	}

lorenzleutgeb:
I am happy to remove restriction (1.) and provide a PR.

Note that (3.) is [discussed here](https://github.com/alpha-asp/Alpha/pull/68#discussion_r147363502). Again, I am happy to remove that restriction and also move the logic to `NaiveGrounder` as suggested.

Can you take care of (2.)? It seems that this is more tightly coupled with the grounder than the other two.

AntoniusW:
Sounds good.

I can take (2.), but this will take some time since it is part of getting the variable-binding dependencies and evaluation order right.

lorenzleutgeb:
Relates to #74 as per the [previous comment](https://github.com/alpha-asp/Alpha/issues/72#issuecomment-340383686).

PR #81, opened 2017-11-11
Grounder modularization

lorenzleutgeb:
I resolved (3.) from #72 and refactored `NaiveGrounder` a bit.

It now has a `NogoodRegistry`, a `WorkingMemory` and a `ChoiceRecorder`.

Also, I decided to split `FixedInterpretation` from `Predicate`. All the interpretation classes that correspond to lambdas or methods are now more independent from `Predicate`.

~~I also introduced `SymbolWithRank` which is used for `Predicate` (predicate symbol and its arity), as well as `FunctionSymbol`, which in turn divides into `Functor` for "real" function names and `Constant` for nullary function symbols.~~ (removed after discussion)


AntoniusW:
Now this is bad timing, since I am also in the middle of several changes to the NaiveGrounder in order to deal with variables bindings correctly. I will open another PR soon, then we can decide which one is easier to put on top of the other. You addressed several other issues than I do right now, so maybe it is not that complicated. :-)

lorenzleutgeb:
OK, then I'll just leave this aside for a moment and we consolidate afterwards.

AntoniusW:
With the changes on variable binding merged into master, this should no longer be blocked. Especially the newly introduced `WorkingMemory` and `ChoiceRecorder` look useful.
In light of #87 the `NoGoodRegistry` is not necessary and the IDs of NoGoods could be removed from Alpha altogether (maybe in another PR or directly here).

lorenzleutgeb:
Yes, I will soon update this PR.

lorenzleutgeb:
I updated this branch and think that these changes really improve the "code quality" of the grounder.

@AntoniusW please reconsider.

src/main/java/at/ac/tuwien/kr/alpha/grounder/NaiveGrounder.java
```diff
@@ -94,8 +89,8 @@ public NaiveGrounder(Program program, java.util.function.Predicate<Predicate> fi
 		for (Atom fact : program.getFacts()) {
 			final Predicate predicate = fact.getPredicate();
 
-			// Record predicate
-			adaptWorkingMemoryForPredicate(predicate);
+			// Record interpretation
```
AntoniusW: "predicate." I think.

lorenzleutgeb: Yes. Sorry.

AntoniusW:
One of the tests still fails, would merge otherwise.

lorenzleutgeb:
Oh, this seems to be a side effect after merging #96 ... will fix.

lorenzleutgeb:
I took the liberty to merge this after your positive comment.

AntoniusW:
This was my intention when writing that comment.

PR #83, opened 2017-11-13
Refactor solver

lorenzleutgeb:


lorenzleutgeb:
I am investigating why that one test case fails. Currently I am unable to reproduce it locally.

src/main/java/at/ac/tuwien/kr/alpha/solver/ChoiceManager.java
```diff
@@ -147,13 +189,63 @@ public void updateAssignments() {
...
+		backjumps++;
+		LOGGER.debug("Backjumping to decision level {}.", target);
+
+		// Remove everything above the target level, but keep the target level unchanged.
+		while (assignment.getDecisionLevel() > target) {
+			backtrackFast();
+		}
+	}
+
+	public void backtrackFast() {
```
AntoniusW: What is the difference between `backtrackFast`, `backtrackSlow`, and `backtrack`? When is the one used and when the other? Please put that in some explanatory comment.

lorenzleutgeb: `backtrack` is private and contains what is common to `backtrackSlow` and `backtrackFast`: It sets counters and recomputes active choice points.

`backtrackSlow` will read and return the assignment that resulted in the choice. This is to enable further analysis. It is called in `DefaultSolver.backtrack` where different actions are taken depending on what happened since the assignment was made.

`backtrackFast` just backtracks the assignment without caring about what the particular assignment was.

I will add a similar explanation in the code.

lorenzleutgeb:
These changes came up because I wanted to refresh myself on how the solver works. I realized that our `ChoiceStack` does not really add much value, and gave more power to the `ChoiceManager`. It is now more  involved in backtracking.

lorenzleutgeb:
Slight decrease in coverage (again), otherwise should be fine.

src/main/java/at/ac/tuwien/kr/alpha/solver/NoGoodStoreAlphaRoaming.java
```diff
@@ -171,7 +174,7 @@ private ConflictCause addAndWatch(final NoGood noGood) {
 				}
 			}
 			// Check strong unassigned for nogoods with head.
-			if (noGood.hasHead() && noGood.getHead() != i && (literalEntry == null || literalEntry.getTruth().isMBT())) {
+			if (noGood.hasHead() && i != 0 && (literalEntry == null || literalEntry.getTruth().isMBT())) {
```
AntoniusW: In this file `0` occurs several times as a magic value, I guess denoting the head. There also seems to be a `HEAD` constant. Could it be used here instead of `0`?

lorenzleutgeb: Yes, `HEAD` should be used. I will add that.

AntoniusW:
Ready to merge after conflicts with master are resolved.

lorenzleutgeb:
I went ahead and merged this after your explicit confirmation :)

Issue #74, opened 2017-10-30
Resolve variable-binding dependencies in rules.


AntoniusW:
In order to allow that `X = 1 + Y` is binding for `X` and other issues with variable binding, the dependencies of variable bindings in rules must be resolved. These dependencies dictate possible evaluation orders of the lazy grounding (i.e., variables restrict the join order of atoms).

We also have a kind of rules that lack an ordinary positive atom with variables to start the usual grounding (e.g., fully ground rules, rules with only interval terms or externals with output variables). Those rules must be treated once, e.g., directly at the beginning, similar to facts.

AntoniusW:
Resolved with #88 merged.

Issue #69, opened 2017-10-27
Represent builtin-atom with its own class.


AntoniusW:
Currently, the builtin atoms (according to the standard) are internally represented by an `ExternalAtom`. I think it would be better to represent them by their own atom class, e.g. `BuiltinAtom` or `ComparisonAtom`. There are two reasons: first, it feels more natural than representing them with `ExternalAtom`, and second, these builtin atoms can bind variables (according to the standard) but it depends on the actual instance of the atom. If the comparison operator is `=` and on one side there is a single variable, then this variable is binding. So, `X = Y + 3` would have `X` binding and `Y` non-binding while for `X +1 = Y` it would be the other way round. I think this behaviour cannot be modelled with `ExternalAtom` in a nice way.

Are there any arguments against introducing this new atom class?

lorenzleutgeb:
No, I think that's a good idea. my only concern is that we might need code very similar to `BuiltinBiPredicate`. Let's see how it goes.

AntoniusW:
Yes, this would supersede `BuiltinBiPredicate` and the new class would use some of its code.

lorenzleutgeb:
OK

AntoniusW:
Resolved with #88 merged.

PR #88, opened 2017-11-15
Fixing variable binding / rule grounding order.

AntoniusW:
Includes grounder refactoring, new builtin atoms following #69, and early arithmetics (not yet standard conformant). Resolves issue #74.


AntoniusW:
My local checkstyle seems to be more permissive than the one running on Travis CI. Any idea on what is wrong here? @lorenzleutgeb

lorenzleutgeb:
> My local checkstyle seems to be more permissive than the one running on Travis CI. Any idea on what is wrong here?

No, not really. The Checkstyle configuration is checked in to Git. Also I would not expect that having a different version installed (which I think is possible since we are using the Checkstyle plugin for Gradle and do not specify a version to use explicitly) has an impact.

src/main/java/at/ac/tuwien/kr/alpha/common/BinaryOperator.java
```diff
@@ -1,8 +1,5 @@
 package at.ac.tuwien.kr.alpha.common;
 
-import at.ac.tuwien.kr.alpha.common.predicates.BuiltinBiPredicate;
-import at.ac.tuwien.kr.alpha.common.predicates.FixedInterpretationPredicate;
-
 public enum BinaryOperator {
```
lorenzleutgeb: Now that we also have `ArithmeticOperator` I think that `BinaryOperator` is confusing and it should be renamed to `ComparisonOperator`.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/BuiltinAtom.java
```diff
@@ -0,0 +1,222 @@
...
+				// Ground term is another term (constant, or function term).
+				resultTerm = groundTerm;
+			}
+			Substitution extendedSubstitution = new Substitution(partialSubstitution);
+			extendedSubstitution.put(variable, resultTerm);
+			return Collections.singletonList(extendedSubstitution);
+		}
+
+		// Treat all other cases, i.e., this is just comparison with all variables bound by partialSubstitution.
+		if (compare(terms.get(0).substitute(partialSubstitution), terms.get(1).substitute(partialSubstitution))) {
```
lorenzleutgeb: Just a matter of style, so you might as well disregard this comment. I like to have they "easy"/"simple" cases of a function at the top. Fox example, if you would move these two cases to the top and wrap them inside an `if (!isLeftAssigning() && !isRightAssigning())` the whole block above (which is more "complicated") would have one less indentation.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/BuiltinAtom.java
```diff
@@ -0,0 +1,222 @@
...
+			return Collections.singletonList(partialSubstitution);
+		} else {
+			return Collections.emptyList();
+		}
+	}
+
+	private boolean compare(Term x, Term y) {
```
lorenzleutgeb: I think that the interpretation should be more separate from the atom itself.

src/main/java/at/ac/tuwien/kr/alpha/grounder/RuleGroundingOrder.java
```diff
@@ -0,0 +1,165 @@
...
+import at.ac.tuwien.kr.alpha.common.atoms.Literal;
+import at.ac.tuwien.kr.alpha.common.terms.VariableTerm;
+
+import java.util.*;
+
+/**
+ * Provides the grounder with information on the order to ground the literals in the body of a rule.
+ * Copyright (c) 2017, the Alpha Team.
+ */
+public class RuleGroundingOrder {
```
lorenzleutgeb: It would be nice to have a few sentences describing how the order is actually established. Just by looking at the implementation I am having a hard time to reconstruct the big picture.

src/main/java/at/ac/tuwien/kr/alpha/grounder/atoms/IntervalAtom.java
```diff
@@ -1,43 +1,70 @@
...
+ *
+ * The first term of this atom is an IntervalTerm while the second term is any Term (if it is a VariableTerm, this will
+ * bind to all elements of the interval, otherwise it is a simple check whether the Term is a ConstantTerm<Integer>
+ * with the Integer being inside the interval.
  * Copyright (c) 2017, the Alpha Team.
  */
-public class IntervalAtom implements Literal {
-	public static final Predicate INTERVAL_PREDICATE = new Predicate("_interval", 2, true);
+public class IntervalAtom extends FixedInterpretationAtom {
+	private static final FixedInterpretationPredicate INTERVAL_PREDICATE = new FixedInterpretationPredicate("_interval", 2, true) {
```
lorenzleutgeb: Sorry, but I am really confused by this. You anonymously create a `FixedInterpretationPredicate`, which has exactly one public method, and this public method will throw because it is not supported. Please explain.

src/main/java/at/ac/tuwien/kr/alpha/grounder/transformation/VariableEqualityRemoval.java
```diff
@@ -0,0 +1,106 @@
...
+		}
+	}
+
+	private void findAndReplaceVariableEquality(Rule rule) {
+		// Collect all equal variables.
+		HashMap<VariableTerm, HashSet<VariableTerm>> variableToEqualVariables = new HashMap<>();
+		//HashSet<VariableTerm> equalVariables = new LinkedHashSet<>();
+		HashSet<Literal> equalitiesToRemove = new HashSet<>();
+		for (Literal literal : rule.getBody()) {
+			if (literal instanceof BuiltinAtom) {
```
lorenzleutgeb:
```java
if (!(literal instanceof BuiltinAtom)) {
    continue;
}
```

Would save you one indentation for the interesting part of the code.

src/test/java/at/ac/tuwien/kr/alpha/AlphaTest.java
```diff
@@ -126,6 +127,8 @@ public void smallGraphWithWrongType() throws Exception {
 	}
 
 	@Test
+	@Ignore
```
lorenzleutgeb: Actually, `@Ignore` is best used with a reason, which you give as a comment. Better:

```java
@Ignore("test program is not safe (external lacking output variables). This should throw some exception.")
```

The same applies to other occurrences below.

src/test/java/at/ac/tuwien/kr/alpha/solver/SolverTests.java
```diff
@@ -592,6 +603,23 @@ public void doubleChoiceRule() throws IOException {
...
 	@Test
+	public void simpleArithmetics() throws IOException {
+		assertAnswerSet("eight(X) :- X = 4 + 5 - 1." +
+			"three(X) :- X = Z, Y = 1..10, Z = Y / 3, Z > 2, Z < 4.",
+
+			"eight(8), three(3)");
+	}
+
+	@Test
+	@Ignore
```
lorenzleutgeb: `@Ignore("with current parser yields wrong result (9)")`

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/BuiltinAtom.java
```diff
@@ -0,0 +1,222 @@
...
+import at.ac.tuwien.kr.alpha.common.terms.VariableTerm;
+import at.ac.tuwien.kr.alpha.grounder.Substitution;
+
+import java.util.*;
+
+/**
+ * Represents a builtin atom according to the standard.
+ * Copyright (c) 2017, the Alpha Team.
+ */
+public class BuiltinAtom extends FixedInterpretationAtom {
```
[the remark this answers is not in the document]
AntoniusW: I think the main problem is that the terms of such fixed interpretation predicates still have an influence on the grounding besides just yielding other substitutions. In the case of builtin equality they also influence the order in which a rule may be grounded and they influence whether a rule is safe or not by sometimes yielding bindings for variables.

Examples: For `X = a` the equality binds the variable `X`, while for `f(X) = f(a)` it binds no variable and for `X +1 = Y` it binds `Y`.

So the predicate alone does not have enough information to determine the `getBindingVariables`. This was my main motivation to put the whole treatment into the atom instead of the predicate. On the other hand, of course, the evaluation at the predicate is now mostly useless since only the one at the atom is used and only in case of the external, it relays to the predicate. So I guess, we could also get rid of the evaluation at the predicate in my approach?

But I also start to see that the other approach (splitting variable binding from the evaluation, then move the evaluation away from the atom) might work. However, I see no clear picture for that at the moment. I am not against trying this direction. What benefits besides `Predicate` just being a symbol could we get?

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/BuiltinAtom.java
AntoniusW: Good point. Thanks!

src/main/java/at/ac/tuwien/kr/alpha/grounder/atoms/IntervalAtom.java
AntoniusW: This is why I suspect that we can get rid of the evaluation at the `Predicate`. The `FixedInterpretationPredicate` is probably not really necessary any longer. At an intermediate stage it was needed, but I think it should now work with an ordinary predicate too and this is just cruft.

src/test/java/at/ac/tuwien/kr/alpha/AlphaTest.java
AntoniusW: Ah, I didn't know that.

AntoniusW:
I will address all your minor comments soon and we should decide how to proceed with the splitting of variable binding / atom representation / fixed evaluation. My solution here was to clamp everything together in the atom representation, because I thought it was necessary (and organized). What organization would you suggest and shall I try that here or do you want to try it in the other PR after rebasing it onto this?

lorenzleutgeb:
If you agree, I would split predicates from their interpretation as I have done in #81:

 * [`FixedInterpretationPredicate implements Predicate` gets removed](https://github.com/alpha-asp/Alpha/pull/81/commits/44e1942f8ccbbf2825ccf0b1b9ee94b0e2300747#diff-da9ef38e819dcffb8b63e3f4be7962e0L11)
 * [`FixedInterpretation` (now standalone) is added](https://github.com/alpha-asp/Alpha/pull/81/commits/44e1942f8ccbbf2825ccf0b1b9ee94b0e2300747#diff-86334115544d8dbd5d16c2c753855eefR11)

Then, the mapping between `Predicate` and `Interpretation` is not obvious anymore (since an interpretation is not a predicate anymore), so [an `ExternalAtom` refers to a predicate *and* the interpretation to apply](https://github.com/alpha-asp/Alpha/pull/81/commits/44e1942f8ccbbf2825ccf0b1b9ee94b0e2300747#diff-f2fe9fefc563c30dce64f4ad2401831aR24).

If I run into trouble with the new binding order related code, I will need your help.

I suggest the following: After you have pushed adaptions regarding the minor comments, please leave a comment that you are done. I will then (in a branch called `variable_binding_split` branch) adapt the changes stated above, and propose a pull-request from the `variable_binding_split` branch into `variable_binding` (this branch). This way I will not pollute your PR and we can fold back the changes easily once we agree that my proposal works out.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/BuiltinAtom.java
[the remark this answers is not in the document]
lorenzleutgeb: I mainly wanted to have this distinction since externals get confusing if we do not split predicate and its interpretation. For example, we have no nice way of representing the arity of both input and output terms.

lorenzleutgeb:
Looks good to me! Maybe you can have a look at #90 and if you like it merge that one, so we have the changes in here.

Issue #89, opened 2017-11-15
Enable Arithmetics


AntoniusW:
In order to realize arithmetics as per the standard, the following is required:

- [ ] The parser needs to take operator precedence into account (`*` and `/` before `+` and `-`).
- [ ] Evaluation of all ground arithmetic terms occurring anywhere as a subterm ( a fact `p(f(g(3+4,a)).` must yield `p(f(g(7,a))` in all answer sets).
- [ ] Ignoring instances of undefined arithmetics ( `p(3+a).` and `p(3/0).` are simply ignored, likewise a constraint `:- not p(3/0).` is also ignored).

votacom:
It also seems that arithmetics with variables is unsupported.

Steps to reproduce:

```
#> cat test.asp
h :- b(1-Y), d(Y).

#> java -jar alpha.jar -i test.asp
Exception in thread "main" java.lang.UnsupportedOperationException: Unsupported syntax encountered: 1-Y
[...]
```

Release v0.2.0

votacom:
I am sorry, my comment was too specific. Apparently, Alpha v0.2.0 does not yet support arithmetics at all.

AntoniusW:
The current master branch supports a little arithmetics, so Alpha should process it if you build from source (see https://github.com/alpha-asp/Alpha#building) and then write your rule as:

`h :- b(B1), B1 = Y-1, d(Y).`

Otherwise arithmetics is still very limited as this issue indicates.

rtaupe:
Apparently it can happen that Alpha does not ground rules in which unsupported forms of arithmetics occur, but no error message is returned to the user, who may not notice in this case that anything is wrong. Can be reproduced with [this program](https://github.com/alpha-asp/Alpha/files/3423139/pup.alpha_clingoheu_20190722.asp.txt), in which the term `O-1` in the following rule seems to be problematic:
`h_assign_1(U1,T,X,W) :- order(X,T,O), assign(U,T1,Y), order(Y,T1,O-1), maxOrder(M), maxUnit(MU), comUnit(U1), U1<=U, maxElem(ME), W=10*MU*(M-O)+2*(ME-X)+U1.`

So as a first step, Alpha should at least report to the user if unsupported arithmetics occur in the input.

AntoniusW:
[...]

Issue #76, opened 2017-10-30
Native lists via external atoms.


AntoniusW:
Lists are very convenient constructs in logic programming and some ASP solvers support them natively. Since our constants can be actual Java objects, it might be easy to allow list construction, membership testing, appending elements, etc. via external atoms that directly work on Java lists.

In a later iteration, the usual list constructs may be directly supported and internally rewritten into the respective external atoms.

lorenzleutgeb:
Just as an inspiration:

 * [Lists in the DLV User Manual](http://www.dlvsystem.com/html/DLV_User_Manual.html#BUILTIN-LIST) defines ~13 built-in predicates (all align well with `java.util.List`).
 * [An ASP System with Functions, Lists, and Sets](https://link.springer.com/chapter/10.1007/978-3-642-04238-6_46) ([DLV-Complex](https://www.mat.unical.it/dlv-complex))

Some conclusions in [Answer Set Programming: A Primer](http://www.kr.tuwien.ac.at/staff/tkren/pub/2009/rw2009-asp.pdf):
> While the theory of ASP is well-developed and applications are expanding, the deployment of ASP to an industrial scale needs further efforts ([cf. Nicola Leone’s talk at LPNMR 2007](lpnmr2007.googlepages.com/nicola-lpnmr07.pdf)). Next generation answer set solvers must be developed which provide better support for the needs in practice.
>
[...]

PR #90, opened 2017-11-15
Split Predicate and FixedInterpretation

lorenzleutgeb:


src/main/java/at/ac/tuwien/kr/alpha/common/atoms/BasicAtom.java
```diff
@@ -44,14 +44,14 @@ public BasicAtom(Predicate predicate, List<Term> terms) {
...
 	}
 
-	public BasicAtom(BasicAtom clone) {
-		this(clone, clone.negated);
+	public BasicAtom(Literal clone) {
+		this(clone, clone.isNegated());
 	}
 
-	public BasicAtom(BasicAtom clone, boolean negated) {
+	public BasicAtom(Atom clone, boolean negated) {
```
AntoniusW: This looks very dangerous, because you can then call `new BasicAtom(new BuiltinAtom(...))` and you get a very weird result.

lorenzleutgeb:
> In my understanding, a predicate in logic is a "predicate symbol" plus an "arity", but it is no symbol itself.

So you suggest that it should not be inside a package named "symbols"? Well, then we just move it out of there or ditch the "symbols" sub-package completely. What I wanted to capture with `SymbolAndRank` (and also with `Predicate` being a subtype) is [from "Techniques for Efficient Lazy-Grounding ASP Solving"]:
> Let C be a finite set of constants, V be a set of variables, and P be a finite
set of predicates with associated arities, i.e., elements of P are of the form p/k
where p is the predicate name and k its arity.

This is what `Symbol` tries to capture. Maybe I should have called it `NameWithArity` or something better, but the essence is the same: It has a "symbol" or "name" component and a "rank" or "arity" component to it. I thought about [Ranked Alphabets](https://en.wikipedia.org/wiki/Ranked_alphabet) when I wrote these classes.

> What is the `Functor` used for? By the way, the `FunctionSymbol` is a symbol with rank while usually a function symbol is just the symbol without arity.

A `Functor` is (as noted in ASP-Core-2) a function name. I introduced it as a subtype of `FunctionSymbol`, to allow for a distinction between `ConstantTerm` which is a `FunctionSymbol` of arity 0 and a `Functor` which normally has an arity of at least 1 and is used in `FunctionTerm`. From your comment:

> By the way, the FunctionSymbol is a symbol with rank while usually a function symbol is just the symbol without arity.

I deduce that we are not really interested in "tracking" the arity of function symbols. I did not know that. I am used to functions of a specific arity, and this is also mentioned in ASP-Core-2.

So, my attempt to structure these things (which were previously not so structured in the code) completely backfired and now I have to justify why it is like it is ... :disappointed: I will try to make the best out of it so that we arrive at a consistent picture. If you feel that we should ditch these changes alltogether, please tell me right away.

My picture is something like this:

 1. Atoms are comprised of a predicate and a list of terms.
 2. A *predicate* has a name and an arity.
 3. A *term* can be either a *variable*, or a *function term*.
 4. A *function term* is comprised of a *function* and a *list of terms*.
 5. A *function* has a name and an arity.
 7. A 0-ary function is a *constant*.
 8. A term consisting of a constant and an empty list of terms is called a constant term. (Note that the notions of a constant and a constant term coincide in our code, which I think is fine).
 9. From (2.) and (5.) I generalize to a "symbol with arity" (`Symbol`).
 10. To separate predicates and functions I introduce the types `Predicate` and `FunctionSymbol`.
 8. To distinguish between functions used in function terms that are not constant terms and constants I introduce `Functor` and `ConstantTerm`. Note that `Functor` is a pure convenience.

And again, my intention was to make this more clear in the code, so that in the end, the picture is clear for anyone that reads the code. Turns out it is not so clear at all...

> We need not adapt this oracle-functions terminology, but I would like one that cannot be easily confused with something so different.

Could you please propose a name? It seems I am not capable at naming these things ... In my view, `FixedInterpretation` is not a complete interpretation in the sense that you describe, but an interpretation function for a predicate. That's why I still don't find the name confusing.

> The `ComparisonOperator` comes with a `FixedEvaluation` which switches on the `ComparisonOperator` received at construction to know what to do. I don't see the rationale behind that spaghetti.

I feel offended. However, again with the whole "splitting" topic here my thinking is that I want to separate the representation of the symbols (be it predicates, or the special case of comparison operators) from how they are "interpreted" (I know you do not like this terminology here). Here the interpretation just moved from `ComparisonOperator` (which represents a symbol) to `FixedInterpretation` which represents what the symbol means and therefore how it should be evaluated. For the case of `ComparisonOperator`s I am happy to give that up. In order to remove the spaghetti, we can just make `ComparisonOperator` a class and have it implement/extend `FixedInterpretation`. Note that, your version of the `BuiltinAtom` also switches based on which `ComparisonOperator` it gets at construction, the two approaches are not that different.

lorenzleutgeb:
A reason why I chose `getSymbol` over `getName` is because for cosntant terms this is the function that gets the object represented by the constant. `getName` would be not very intuitive here.

AntoniusW:
I'm sorry to be so picky about the terminology, I know that searching for a good one can be really annoying and time consuming. I have grown to value a clear terminology, because I have seen how confusing a bad one can be. So to me, those discussions and searching for the right words have a worthy goal. On many of your points I strongly agree with you, like splitting the "interpretation" from the "representation" (which is nothing else than semantics and syntax, respectively).

On the other hand, I am also cautious with generalizing everything, since I have seen code where trivial settings were so hidden in layers and layers of abstraction that even simple configurations were very hard to track down (i.e., I had to use a debugger to figure out where some settings came from). This is the reason, why I tend to put things belonging together close to each other (like putting syntax and semantics directly into `BuiltinAtom`).

I think, our main difference is the question what kind of "dimension" to pick first for distinguishing. You would pick the syntax-semantics dimension first and then distinguish on the elements that can appear in an ASP program. My approach was to first distinguish on the elements of ASP and for each element then distinguish on syntax-semantics, if it is necessary (for external atoms it is necessary while for builtin atoms not). Both approaches have their merits and I currently perceive the have-things-together advantage of the latter as a bit more valuable. But I am open to change my mind on this issue, if you hand me some arguments. Maybe I am also completely misunderstanding your approach, in this case, please help me.

Now, on to the "minor" terminology issues:
- Symbols with rank is, as far as I know, the more general name while arity is more specific to CS and logic. Since we don't need the general thing, it is better to be more specific.
- Having function symbols and predicate symbols associated with a fixed arity is convenient in logic and scientific writing, but for logic programming this is most often relaxed. For example, 
`p(f(a,f(b))). p(1,2).` is usually a valid ASP program, although there is not just one arity for `f` and neither is there for `p`. So a predicate is a "predicate symbol plus arity" and not a mapping from predicate symbols to arities. For function terms there usually is no necessity to store the arity at all, for predicates it is needed to make the look-up in the correct storage since p/1 and p/2 are usually two different predicates stored in different locations.
- Your picture of things and my view:
1. Agree.
2. Agree.
3. A term can be a variable, function term, or constant term. (Sometimes constants are seen as 0-ary functions, but most often they are different).
4. Agree.
5. Not agree, although it is possible to view it a such. In the ASP literature constants are most often their own type of objects. This may have to do with the grounding, where after the grounding (if you get a finite one) you can simply replace every function term with a fresh constant and then you have only constants and need no function terms at all. Now, Alpha is not constrained to constants and we could indeed market this much more. Since we have a lot of other hot topics, I feel that "just constants" make the presentation in papers easier. Hence, I want them separate from functions.
7. Mostly agree. By the above, there cannot be a constant with a list of terms. Note that also the standard calls `f()` a synonym for the constant `f`.
8. I agree that for 2-5 one can compute an arity, but there is no need to store the arity (other than for predicates). Specifically, there is no functional relation between symbols and arities.
9. I would stick to `FunctionTerm` to be clear (or `terms.Function`) and a `FunctionTerm` is comprised of a function symbol and a list of terms.
10. I think `Functor` is a not a good name and I guess in the standard they were aware of that since there it is immediately followed by an explanation that this is the function name. To me, a `Functor` is a function symbol.

I start to think that an UML diagram could be helpful, again.

- Naming the external evaluation class: I have the same problems as you with finding a good name. I see where `FixedInterpretation` is fitting, but I fear it can be mistaken for an interpretation of more than just the one atom (as I did the moment I saw it). I have no good name yet, but often through discussion one can find a suitable one. Since it is semantics, interpretation is definitely not wrong, so maybe `FixedPartialInterpretation` would be better. On the other hand, it is not any partial interpretation but tailored to a specific predicate. Maybe `FixedPredicateInterpretation`? We could also argue that this is semantics and go with `FixedPredicateSemantics` or `FixedSemanticsOracle` or `FixedSemanticsPredicateEvaluator`. Do you have any other ideas on this?

I did not intend to offend you, sorry. Much to the contrary, I highly appreciate your drive to structure the code well! Your contribution here is _really_ valuable!

- Regarding `BuiltinAtom`, my view on that indeed is quite narrow on just the thing according to the standard, so comparison and nothing else. Until now I saw no real reason to think of it as a wider concept. The native lists could be such a wider concept, but I am not fully sure since I did not think it through. Lists are convenient to use at positions where terms occur (like `pred( foo, [H | T])` ). So if one goes into that direction it would make sense to have a richer input language and rewrite that into external atoms that power the native lists. How would you fit lists into the picture?

lorenzleutgeb:
OK. I have pedalled back a bit.

 * `SymbolWithArity` and `Symbol` are both gone.
 * `Predicate` still is separated from its interpretation, but it is not a `SymbolWithArity` anymore so it has `getName` and `getArity`.
 * `Functor` is gone and `FunctionTerm` has a `String symbol` now that just contains the the function symbol associated with the term.
 * I renamed `FixedInterpretation`s to `PredicateInterpretation`s.

Some built-in predicates (that would probably clash in `BuiltinAtom`, therefore we favor more specific `Atom`/`Literal` implementations):

 * `<, >, <=, >=, =, !=` called comparison predicates, we already have them.
 * `#int, #succ, #prec, #mod, #absdiff, #rand, +, *, -, /` called arithmetic predicates, you already started implementing them. See #89 
 * `#append, #delnth, #flatten, getnth, #head, #insLast, #insnth, #last, #length, #member, #reverse, #subList, #tail` called list predicates. See #76

AntoniusW:
How do you feel about calling `PredicateInterpretation` an `InvariablePredicateInterpretation` or `PredicateInterpretationInvariable`?
This fixed/invariable does not need to propagate through to all subcasses and it sounds more fitting to have some of the fixed-association verbally remaining in places like the `NaiveGrounder`.

PR #96, opened 2017-11-18
Literate Programming

lorenzleutgeb:
While reading some ASP code today I realized that the authors had quite lengthy documentation blocks as comments above predicate definitions or smaller chunks of code. I thought that [Literate Programming](https://en.wikipedia.org/wiki/Literate_programming) would be the preferred layout to write code like this, and we might want to support it.

For example, Clique in Literate Answer Set Programming (with disjunction):

```markdown
# The *3-Clique Problem*

The *3-Clique Problem* is a decision problem. Given a *graph*, consisting of *vertices*
and *edges* that connect two vertices each, can we find 3 vertices, such that they are
all pairwise connected?

## Instance

We will be looking at four nodes

    vertex(1..4).

that are connected as follows:

    edge(1,2).
    edge(1,3).
    edge(1,4).
    edge(2,3).

To ensure undirectedness of edges, we add the following rule to establish symmetry:

    edge(U,V) :- edge(V,U).

## Guess

We will use the `clique` predicate to mark those vertices that are in the clique.
For any vertex, it might or might not be in the clique.

    clique(V) v -clique(V) :- vertex(V).

## Check

At least k vertices must be in the clique.

    :- #count{V : clique(V)} < 3.

Every two vertices that are in the clique must be connected through an edge.

    :- clique(U), clique(V), U != V, not edge(U,V), not edge(V,U).
```


AntoniusW:
That is a big-step enhancement and quite far ahead of all the other issues. I did not recall Literate Programming but I've had related ideas a long time ago (and with no connection to ASP). This can be some very nice addition to the system. I need some more time to fully understand its implication, however.

For example, you changed the parser to ignore all errors and continue with the rest, right? So if I write a comment like `This is the code about bovines (cows). For whatever reasons...` then the answer sets will contain `bovines(cows)` because that looks like a fact? This seems to be dangerous.

On the other hand, I like the idea that your example above is the actual input!

lorenzleutgeb:
No, I wrote a small pre-processing routine `Util.literate` which operates on a stream of lines `Stream<String>`. It comments out everything that is not indented by at least four spaces and removes those spaces in case they are there.

So what the parser would see in case of the above example is the following:

```markdown
% # The *3-Clique Problem*
% 
% The *3-Clique Problem* is a decision problem. Given a *graph*, consisting of *vertices*
% and *edges* that connect two vertices each, can we find 3 vertices, such that they are
% all pairwise connected?
% 
% ## Instance
% 
% We will be looking at four nodes
% 
vertex(1..4).
% 
% that are connected as follows:
% 
edge(1,2).
edge(1,3).
edge(1,4).
edge(2,3).
% 
% To ensure undirectedness of edges, we add the following rule to establish symmetry:
% 
edge(U,V) :- edge(V,U).
% 
% ## Guess
% 
% We will use the `clique` predicate to mark those vertices that are in the clique.
% For any vertex, it might or might not be in the clique.
% 
clique(V) v -clique(V) :- vertex(V).
% 
% ## Check
% 
% At least k vertices must be in the clique.
% 
:- #count{V : clique(V)} < 3.
% 
% Every two vertices that are in the clique must be connected through an edge.
% 
:- clique(U), clique(V), U != V, not edge(U,V), not edge(V,U).
```

Note that this preserves line numbers.

There might be more sophisticated implementation using Lexer Modes with ANTLR, but I did not get that to work.

rtaupe:
Interesting idea! It reminds me of [Achievements in Answer Set Programming](https://arxiv.org/pdf/1608.08144.pdf), which seems to be a related concept.

lorenzleutgeb:
I think that Literate Programming and Achievements in Answer Set Programming can be combined, and this does not look too bad.
 IMO Achievements is the postfix version of what I did naturally in my example (i.e., instead of first explaining the rule and then introducing it, achievements record the effect of the rule after stating it). I slightly adopted Fig. 4 from your reference and got the following:

```markdown
# NQueens

input: positive integer n (size of the board).

## Instance

A square is represented as a pair, column number and row number,
both from the set `{1,..,n}`.

    row(1..n).

*achieved*: `row/1 = {1,...,n}`.

    col(1..n).

*achieved*: `col/1 = {1,...,n}`.

## Guess

    n { queen(I,J) : col(I), row(J) } n.

*achieved*: Set `queen/2` consists of n squares.

## Check

    :- queen(I,J), queen(I,JJ), J != JJ.

*achieved*: Each column includes at most one square from `queen/2`.

    :- queen(I,J), queen(II,J), I != II.

*achieved*: Each row includes at most one square from `queen/2`.

    :- queen(I,J), queen(II,JJ), (I,J) != (II,JJ), |I-II|=|J-JJ|.

*achieved*: Each diagonal includes at most one square from `queen/2`.
```

PR #105, opened 2017-11-27
Reduce assignment branch condition size

rtaupe:
Reduce assignment branch condition size (reported by Codebeat)

### DIFF
--- a/src/main/java/at/ac/tuwien/kr/alpha/solver/NoGoodStoreAlphaRoaming.java
+++ b/src/main/java/at/ac/tuwien/kr/alpha/solver/NoGoodStoreAlphaRoaming.java
@@ -369,21 +369,33 @@ public class NoGoodStoreAlphaRoaming implements NoGoodStore, Checkable {
 		Watches<BinaryWatch, WatchedNoGood> watchesOfAssignedAtom = watches(entry.getAtom());
 
 		// Process binary watches.
-		for (BinaryWatch binaryWatch : watchesOfAssignedAtom.binary.getOrdinary(entry.getTruth().toBoolean())) {
-			ConflictCause conflictCause = assignWeakComplement(binaryWatch.getOtherLiteralIndex(), binaryWatch.getNoGood(), entry.getDecisionLevel());
+		Set<BinaryWatch> ordinaryBinaryWatches = watchesOfAssignedAtom.binary.getOrdinary(entry.getTruth().toBoolean());
+		ConflictCause conflictCause = processBinaryWatchesWeakly(ordinaryBinaryWatches, entry.getDecisionLevel());
+		if (conflictCause != null) {
+			return conflictCause;
+		}
+
+		// Check all watched multi-ary NoGoods.
+		Set<WatchedNoGood> ordinaryMultaryWatches = watchesOfAssignedAtom.multary.getOrdinary(entry.getTruth().toBoolean());
+		return processMultaryWatchesWeakly(ordinaryMultaryWatches, entry.getWeakDecisionLevel(), entry.getLiteral());
+	}
+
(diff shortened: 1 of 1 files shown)


AntoniusW:
Making the `NoGoodStoreAlphaRoaming` easier to understand is a good idea and I fully support it. If I understand correctly, then you move two loops into their own methods, right? Personally, I don't like this change, because it a) deviates more from the paper and b) feels to me like it is adding complexity by introducing another layer of indirection. My informal reading of the previous code is "get all binary watches, loop on them and abort if one conflicts", now it seems to me "get all binary watches, hand them to the processor and abort if the processors conflicts. In the processor, loop over all watches and abort if one conflicts." What were your reasons in favor of this change? Again, I'd like to improve/simplify this nogood store.

rtaupe:
I agree that the return statements within the loops pose a problem to the refactoring, and maybe the code is now really harder to understand than before. Unfortunately, I saw no way to split the method up without handling returns the way they are handled now.

The general reason for my change was that the methods `propagateWeakly` and `propagateStrongly` were, in my opinion, overly long and did not follow the [single responsibility principle](https://en.wikipedia.org/wiki/Single_responsibility_principle). Also, I thought that this could be a first step after which we could more easily share code between the two methods, i.e. implement the basic structure in an (abstract) base class that is implemented once for weak and once for strong propagation.
The concrete event because of which I am proposing the change is that codebeat has been criticising the high assignment branch condition size in `propagateWeakly` in #103.

AntoniusW:
There are multiple reasons, why the code is so complex. First, the 3WL propagation that this class implements is no trivial algorithm. Second, this is most likely the hot spot of the solver, so I tried to avoid unnecessary things like iterating twice over the nogood, i.e., it collects all data on one go. This unfortunately mingles semantically unrelated checks.

Maybe one could extract several methods that each return more than one value and then hope for the JIT compiler to inline every one of them again? Before committing that, however, I would like to see some benchmarks showing that indeed performance does not degenerate (much).

Regarding single responsibility: What are the multiple reasons for change in `propagateWeakly` and `propagateStrongly`?
Also, I currently see no clear benefit in creating a common base class, since both propagations are very different in their innermost workings (weak uses 2 watches, while strong only 1 and some consistency checks are not needed for strong propagation because they are already triggered by weak propagation).

rtaupe:
One reason for change is when something in the processing of binary watches changes, another is when something in the processing of multary watches changes, that´s why I have split the method that way. For sure there must be more reasons, otherwise the method would not be so long.
I have not looked at the algorithm in the paper, but only at the implementation. For me, comments describing what the following block of code does are always an indication that this block can be extracted into a new method to decrease method size.

If you think the implementation deviates more from the paper with the proposed changes, let´s just forget this pull request and find a better change that aligns with the paper. Unfortunately, I do not quite understand your suggestion:
> Maybe one could extract several methods that each return more than one value and then hope for the JIT compiler to inline every one of them again?

AntoniusW:
I tried some refactoring in 7e2d7c2 (committed directly to the other branch/PR), following your ideas to break the weak propagation into smaller methods. I just manually checked that the JVM indeed starts to inline the introduced methods (although they are quite close to not being inlined because of their size).

lorenzleutgeb:
Out of curiosity: How did you verify that these methods are indeed inlined?

AntoniusW:
Running java (Oracle JVM) with the following command line options will print a lot of JIT information:
`-XX:+PrintCompilation -XX:+UnlockDiagnosticVMOptions -XX:+PrintInlining`

There are several inlining options, mainly due to small methods and due to some methods being in the hot path. Those methods for propagation are too large to be inlined directly but they are inlined after becoming hot (although they are almost too large for that even).

Issue #42, opened 2017-01-27
DefaultSolver / NaiveHeuristic: nondeterministic behaviour and wrong solution


rtaupe:
When calling `at.ac.tuwien.kr.alpha.solver.HanoiTowerTest.testSimple()` (on [testing-hanoi](https://github.com/AntoniusW/Alpha/tree/testing-hanoi) branch) with `DefaultSolver` and deterministic `NaiveHeuristic`, the following can be experienced:

1. Nondeterministic behaviour: Sometimes the solver finds an answer set quickly and returns it (after 87 decisions), sometimes it takes long time and finds no answer set (after 34228 decisions).
2. In the latter case, the result is actually wrong, because there is a solution to the problem.

AntoniusW:
The faulty behaviour seems to no longer occur in the current master branch and nondeterminism has also been removed recently. Please verify.

rtaupe:
Sorry, cannot confirm.
On my system, the bug still occurs, buth on the current master branch and on  [testing-hanoi](https://github.com/AntoniusW/Alpha/tree/testing-hanoi) after rebasing it on master.
Just the numbers have changed: When successful, the answer set is found after 7086 choices; when unsuccessful, 42207 choices are done.

rtaupe:
**Update:** testing-hanoi branch has been rebased on master. The problem still occurs in the _default/naive/alpharoaming/NAIVE_ configuration (but after executinig the test case a few times it seems that the wrong result is now produced always, i.e. nondeterminism in this respect seems to be gone).

rtaupe:
It seems that I could isolate the problem a little: When adding either of the facts `noWhere(0,2).`, `where(0,3).` or `diskWhere(0).` to the input, the problem becomes quickly solvable and a correct answer set is found, so the solver seems to have a problem in deriving `noWhere(0,2).` or `where(0,3).` (the third fact can directly be derived from the second one).

AntoniusW:
I am getting progress on this, luckily there are two almost identical runs with different configuration (one failing, one not), after I changed a remaining `HashSet` into a `LinkedHashSet` in the grounder. The missing answer set then seems to be caused by the nogood store accidentally forgetting to set the alpha watch when a nogood with certain properties is added. I should have a fix later.

AntoniusW:
Merging #109 locally into `testing-hanoi` should fix the issue with `testSimple()` and also remove another source of non-determinism. Please check if this indeed fixes all of the problems raised in this issue.

[...]

PR #109, opened 2017-12-01
Issue 42.

AntoniusW:
Addresses #42.


AntoniusW:
Test case added, but since the watches are a private matter of the nogood store this is a bit hard to test. The test now checks that under the very specific conditions of this issue, the watch seems to be there. As a safeguard, I already introduced a check at the end of the addition function, ensuring that every nogood with head has an alpha watch set:
https://github.com/alpha-asp/Alpha/blob/9e2b427deea845015229b6b0eeaadefa8a3b1f9c/src/main/java/at/ac/tuwien/kr/alpha/solver/NoGoodStoreAlphaRoaming.java#L273-L275

PR #118, opened 2018-02-08
Refactor Literal to make it contain an Atom instead of subclassing it (and fix minor other issues)

### LINKED ISSUE #117: BasicAtom.substitute(Substitution) does not regard value of negated
`BasicAtom.substitute(Substitution)` does not copy the value of the `negated` member variable to the new `BasicAtom` instance.

rtaupe:
fixes #117, #119, and #120


rtaupe:
don´t merge yet, apparently this causes problems not detected by our test suite

AntoniusW:
Yes, I discovered this issue on a private branch recently, but my fix there (basically the one as proposed) caused other issues, because printing uses negation and, if I remember correctly, somehow atoms in answer sets showed up as being negated. What problems did you encounter?

Maybe this indicates the need for some further refactoring/clarification of the employed classes.

rtaupe:
@AntoniusW yes, basically the same problems (cf. #119). I´ll propose a fix, then we can think about whether a refactoring would be better. Maybe `Literal` shouldn´t extend `Atom`.

rtaupe:
@AntoniusW what do you think of the refactorings in f4bd744a4e5671207c516b1e46d9fa31a6066783?

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/Atom.java
```diff
@@ -20,13 +47,13 @@
 	 * List of all variables occurring in the Atom that are potentially binding, i.e., variables in positive atoms.
 	 * @return
 	 */
-	List<VariableTerm> getBindingVariables();
+	List<VariableTerm> getBindingVariables(boolean negated);
```
AntoniusW: I think this `boolean negated` is a bad idea, because the outside now must know which kind of Atom this is and the rest of the refactoring actually tried to move negation outside of `Atom` (which is a good idea!).

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/Literal.java
```diff
@@ -1,8 +1,93 @@
...
+	 * @see Atom#substitute(Substitution)
+	 */
+	Literal substitute(Substitution substitution);
+
+	/**
+	 * List of all variables occurring in the Atom that are potentially binding, i.e., variables in positive atoms.
+	 * @return
+	 */
+	default List<VariableTerm> getBindingVariables() {
+		return getAtom().getBindingVariables(isNegated());
```
AntoniusW: Here the negation should be used to either return `getAtom().getBindingVariables()` or `Collections.emptyList()`.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/Literal.java
```diff
@@ -1,8 +1,93 @@
...
+	}
+
+	/**
+	 * List of all variables occurring in the Atom that are never binding, not even in positive atoms, e.g., variables in intervals or built-in atoms.
+	 * @return
+	 */
+	default List<VariableTerm> getNonBindingVariables() {
+		return getAtom().getNonBindingVariables(isNegated());
```
AntoniusW: And here we either return `getAtom().getNonBindingVariables()` or the union of the binding and non-binding variables in case the literal is negated.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/Literal.java
rtaupe: You´re absolutely right, I´ll do that.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/Literal.java
rtaupe: You´re absolutely right, I´ll do that.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/Atom.java
```diff
@@ -35,6 +63,22 @@
...
+	default Literal toLiteral() {
+		return toLiteral(true);
+	}
+	
+	/**
+	 * Creates a literal containing this atom which will be negated if {@code positive} is {@code false}
+	 * @param positive
+	 * @return
+	 */
+	default Literal toLiteral(boolean positive) {
```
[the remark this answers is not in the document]
rtaupe: You´re right, we should not have two different meanings attached to this or similar boolean parameters. However, if you just see `toLiteral(true)` without seeing the parameter name, I think it is more natural to interpret this as _to positive literal_ than as _to negated literal_.
Would you prefer `boolean negated` over `boolean positive`?

rtaupe:
@lorenzleutgeb In my opinion, we need `ComparisonLiteral` because its methods `getBindingVariables`, `getNonBindingVariables`, and `getSubstitutions` depend on the `negated` member which does not exist in `Atom`. But actually we should be able to make `ComparisonLiteral` extend `AtomLiteral` instead of having its own `atom` and `negated` members.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/AtomLiteral.java
```diff
@@ -0,0 +1,70 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package at.ac.tuwien.kr.alpha.common.atoms;
+
+import at.ac.tuwien.kr.alpha.grounder.Substitution;
+
+/**
+ * Contains a potentially negated {@link Atom}
+ */
+public class AtomLiteral implements Literal {
```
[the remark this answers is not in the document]
rtaupe: ok, will merge them

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/Literal.java
```diff
@@ -38,35 +38,52 @@
...
+	}
 
-	boolean isNegated();
+	public boolean isNegated() {
+		return negated;
+	}
 
 	/**
 	 * Returns a new copy of this literal whose {@link Literal#isNegated()} status is inverted
 	 * (if the atom cannot be negated, the literal may be returned without creating a copy)
```
AntoniusW: Is the comment in parentheses still correct? It sounds a bit off, since I would rather expect an exception if a literal cannot be negated.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/Literal.java
```diff
@@ -75,9 +92,8 @@
 	 * 
 	 * @return
 	 */
-	default Set<VariableTerm> getNonBindingVariables() {
+	public Set<VariableTerm> getNonBindingVariables() {
 		if (isNegated()) {
-			Atom atom = getAtom();
 			return SetUtils.union(atom.getBindingVariables(), atom.getNonBindingVariables());
 		} else {
 			return Collections.emptySet();
```
AntoniusW: I think it should be `atom.getNonBindingVariables()` here. Or is it guaranteed that each `Literal` has only binding variables? In that case the union in the other if-branch should be removed and replaced with just the binding variables.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/Literal.java
rtaupe: No, not correct. Thanks! Fixed in 68e8f531eb2f4a74bbdab8a632b55912aa3a8f6c

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/Atom.java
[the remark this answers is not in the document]
rtaupe: @AntoniusW what do you think?

rtaupe:
@lorenzleutgeb @AntoniusW
I now got rid of ComparisonLiteral. Now, `getBindingVariables()` and `getNonBindingVariables()` implementations in some `Atom` implementations use an additional `negated` parameter because their results depend on this information (and apparently there is no uniform handling that works for all sorts of atoms and can be implemented in `Literal`). Is this better?

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/Literal.java
rtaupe: Thanks for pointing this out, I actually introduced some issues here. Now, [LiteralBindingNonBindingVariablesTest](https://github.com/alpha-asp/Alpha/blob/issue_117/src/test/java/at/ac/tuwien/kr/alpha/common/atoms/LiteralBindingNonBindingVariablesTest.java) ensures that binding and non-binding variables are the same as [before my changes](https://github.com/alpha-asp/Alpha/blob/test/non-binding_vars/src/test/java/at/ac/tuwien/kr/alpha/common/atoms/AtomBindingNonBindingVariablesTest.java) (see additional elaboration [below](https://github.com/alpha-asp/Alpha/pull/118#issuecomment-381642604)).

AntoniusW:
I guess this is a difficult situation now. The current solution forces all `Atoms` to know about negation or at least know how its variable binding is influenced by negation. On the other hand, the `Literal` cannot decide this in general, since we have the weird functionality that `not X != 3` is actually binding `X` and external atoms also bring their own conditions on variable bindings. So it definitely makes sense that each `Atom` must know how negation influences its variable binding.

On the other hand, if `Atoms` deal with negation anyway, what for do we need the `Literal` then? So I now wonder, whether we should get rid of it (by adding a `boolean negated` to each `Atom` and then renaming every `Atom...` into a `Literal...` so that in the end there are only literals and no more atoms). What do you think? It feels really strange to suggest, but I seriously wonder what counter arguments are there?

rtaupe:
It really would feel strange to get rid of atoms completely. Apart from the fact that atoms and literals are just two different concepts, I think one counter-argument is that we use atoms in many places: e.g., facts are represented by atoms, answer sets consist of atoms, rule heads consist of atoms, ...

However, if we keep atom and literal separate, negation should only occur in literals. So one more suggestion (I think this should work but I haven´t tried it yet): Let´s keep both `Atom` and `Literal`, but introduce additional subclasses of `Literal` (`BasicLiteral`, `ExternalLiteral`, ...) in which we implement `getBindingVariables` and `getNonBindingVariables`. What do you think?

AntoniusW:
Thank you for the arguments in favor of keeping `Atom` and `Literal` separated.

Your suggestion of introducing subclasses of `Literal` and moving the variable binding there looks interesting. It feels a bit unnecessary, but by moving the binding into `Literals` it becomes clear that `Atoms` are building blocks that by themselves cannot occur in a rule directly. I like that.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/Atom.java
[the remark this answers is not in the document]
rtaupe: As just discussed with @AntoniusW offline, we go for `boolean positive` and change all occurances of `boolean negated` globally (but in a separate pull request).

rtaupe:
Done.

I introduced a small inline TODO here: https://github.com/alpha-asp/Alpha/blob/fbc53860d08a3ef358387b96a3f77d9f1248da89/src/main/java/at/ac/tuwien/kr/alpha/grounder/NoGoodGenerator.java#L145-L146
I think this can be resolved when this FIXME is resolved:

https://github.com/alpha-asp/Alpha/blob/fbc53860d08a3ef358387b96a3f77d9f1248da89/src/main/java/at/ac/tuwien/kr/alpha/grounder/NonGroundRule.java#L78-L79

The reason for this is that `NonGroundRule` currently does not consist of literals, but of atoms.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/FixedInterpretationLiteral.java
```diff
@@ -1,14 +1,42 @@
...
 /**
- * Represents an atom whose ground truth value(s) are independent of the current assignment.
- * Examples of such atoms are builtin atoms and external atoms.
- * Copyright (c) 2017, the Alpha Team.
+ * Represents a literal whose ground truth value(s) are independent of the current assignment.
+ * Examples of atoms underlying such literals are builtin atoms and external atoms.
+ * Copyright (c) 2017-2018, the Alpha Team.
  */
-public interface FixedInterpretationLiteral extends Literal {
+public interface FixedInterpretationLiteral {
```
AntoniusW: Is there a reason why `FixedInterpretationLiteral` no longer extends `Literal` (and becomes an abstract class by that)? I'm asking, because this now in principle allows that something is a `FixedInterpretationLiteral` but not a `Literal`.

src/main/java/at/ac/tuwien/kr/alpha/grounder/parser/ParseTreeVisitor.java
```diff
@@ -298,24 +320,24 @@ public Literal visitNaf_literal(ASPCore2Parser.Naf_literalContext ctx) {
 		// naf_literal : NAF? (external_atom | classical_literal | builtin_atom);
 		isCurrentLiteralNegated = ctx.NAF() != null;
```
AntoniusW: It seems that `isCurrentLiteralNegated` now can be a variable local to this method, since it is no longer passed around.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/FixedInterpretationLiteral.java
rtaupe: The only reason is that `Literal` became a class and an interface cannot extend a class. I´m not sure if one of the two design alternatives is better, but since you seem to have a preference towards `FixedInterpretationLiteral` extending the abstract class, I implemented this in e089a5dd3317cf8fc410c0376350d82da3538faf.

src/main/java/at/ac/tuwien/kr/alpha/grounder/parser/ParseTreeVisitor.java
rtaupe: Right, thanks! Implemented in a217ffd89f714badf89a5ba8700cf7da78557084.

rtaupe:
Thank you for your reviews.
@AntoniusW will you merge or shall I?

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/Atom.java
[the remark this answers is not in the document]
rtaupe: https://github.com/alpha-asp/Alpha/pull/129

PR #125, opened 2018-03-19
Improvements to design of branching heuristics interface

rtaupe:
- move `getActivity` method into new interface so it has to be implemented only by heuristics for which the method is relevant
- combine `chooseAtom()` and `chooseSign()` as `chooseLiteral()`
- replace single fallback heuristic by chain of arbitrary length


rtaupe:
@AntoniusW
Thank you for your comment. Unfortunately, I disagree with its first part. `ChainOfBranchingHeuristics` does not simulates a list, but contains one as a member. The class provides four methods, `violatedNoGood`, `analyzedConflict`, `newNoGood`, and `chooseLiteral`, each of which iterates over the elements of the list. If a `List<List<BranchingHeuristic>` were used in `DefaultSolver` directly, the latter would have to do this iteration in every place where it currently just calls one of these methods.

If you still think that `ChainOfBranchingHeuristics` is less elegant, I´ll remove it.

rtaupe:
As just discussed offline between @AntoniusW and me, `ChainOfBranchingHeuristics` will be renamed to `ChainedBranchingHeuristics` and we keep the other changes.

PR #130, opened 2018-05-06
Add justification analysis.

AntoniusW:
Compute justifications for atoms (avoids chronological backtracking when MBT remains at fixpoint).


src/main/java/at/ac/tuwien/kr/alpha/grounder/IndexedInstanceStorage.java
```diff
@@ -168,6 +169,41 @@ public void removeInstance(Instance instance) {
...
+		for (int i = 0; i < atom.getTerms().size(); i++) {
+			Term testTerm = atom.getTerms().get(i);
+			if (testTerm.isGround()) {
+				return i;
+			}
+		}
+		return -1;
+	}
+
+	public Collection<Instance> getInstancesFromPartiallyGroundAtom(Atom substitute) {
```
lorenzleutgeb: This signature combined with the returned expressions in lines 189 and 191 is a little strange. Since `getInstancesMatchingAtPosition` returns `List<Instance>` and `new ArrayList<>(getAllInstances()` will be a `List<Instance>` I would expect the method to return this type. If you prefer to return `Collection<Instance>` please justify this with a comment. Similar reasoning applies to `getInstancesMatching` below.

src/main/java/at/ac/tuwien/kr/alpha/grounder/IndexedInstanceStorage.java
```diff
@@ -168,6 +169,41 @@ public void removeInstance(Instance instance) {
...
+		if (firstGroundTermPosition != -1) {
+			Term firstGroundTerm = substitute.getTerms().get(firstGroundTermPosition);
+			return getInstancesMatchingAtPosition(firstGroundTerm, firstGroundTermPosition);
+		} else {
+			return new ArrayList<>(getAllInstances());
+		}
+	}
+
+	public Collection<Instance> getInstancesMatching(Atom matchAtom) {
+		ArrayList<Instance> matchingInstances = new ArrayList<>();
```
lorenzleutgeb: It is a common convention to have the more general/abstract collection type on the left, e.g.:
```java
List<T> xs = new ArrayList<T>();
```
instead of
```java
ArrayList<T> xs = new ArrayList<T>();
```
and
```java
Set<T> xs = new HashSet<T>();
```
instead of
```java
HashSet<T> xs = new HashSet<T>();
```

The main reason being that this way the code that then actually deals with `xs` must comply with the more general behaviour, and cannot depend on the concrete implementation being used. The implementation to be used is only relevant at initialization time. Of course there are exceptions when one really needs the concrete implementing type throughout, but in most cases you just want a list or a set, and that suffices. The special cases where one depends on the implementation of a collection class is a gain something that should be justified separately in my opinion.

src/main/java/at/ac/tuwien/kr/alpha/solver/ArrayAssignment.java
```diff
@@ -361,6 +362,15 @@ public Entry get(int atom) {
 		return assignment.get(atom);
 	}
 
+	Integer getSomeMBTAssignedAtom(Grounder grounder) {
```
lorenzleutgeb: This is giving me a headache. We now have `Grounder#getAtom` which just forwards to the grounder's `AtomStore`, and then the solver calls `ArrayAssignment#getSomeMBTAssignedAtom` passing in the grounder. Additionally, this method depends on a concrete implementation of `Atom`. Very odd.

It seems that both grounder and solver need to be aware of the assignment. If we would have this then the solver could just ask the grounder for MBT assigned atoms under an assignment. We would have something like: `Integer Grounder#getSomeMBTAssignmentAtom(Assignment)` (cf. `Grounder#getUnassignedAtoms`). That would seem a little more natural to me, and `BasicAtom` would not leak into the assignment (see new import in this file). Maybe I am missing something and this is not possible after all...

src/main/java/at/ac/tuwien/kr/alpha/solver/DefaultSolver.java
```diff
@@ -78,8 +87,11 @@ public DefaultSolver(Grounder grounder, NoGoodStore store, WritableAssignment as
 		this.branchingHeuristic = ChainedBranchingHeuristics.chainOf(
 				BranchingHeuristicFactory.getInstance(branchingHeuristic, grounder, assignment, choiceManager, random),
 				new NaiveHeuristic(choiceManager));
+		this.disableJustifications = disableJustifications;
 	}
 
+	private boolean disableJustificationAfterClosing = true;
```
lorenzleutgeb: Please move this to the top along with all other members.

src/main/java/at/ac/tuwien/kr/alpha/solver/DefaultSolver.java
```diff
@@ -224,15 +231,141 @@ private boolean learnBackjumpAddFromConflict(ConflictCause conflictCause) {
 		return true;
 	}
 
+	private boolean justifyMbtAndBacktrack() {
+		mbtAtFixpoint++;
+		// Run justification if enabled and possible.
+		if (!disableJustifications && grounder instanceof ProgramAnalyzingGrounder) {
```
lorenzleutgeb: You could treat the opposite case (which is considerably simpler) first, and save one indentation for this huge chunk.

src/main/java/at/ac/tuwien/kr/alpha/solver/DefaultSolver.java
```diff
@@ -224,15 +231,141 @@ private boolean learnBackjumpAddFromConflict(ConflictCause conflictCause) {
...
+			if (!backtrack()) {
+				logStats();
+				return false;
+			}
+		}
+		return true;
+	}
+
+	private boolean treatConflictAfterClosing(NoGood violatedNoGood) {
+		if (!disableJustificationAfterClosing && !disableJustifications && grounder instanceof ProgramAnalyzingGrounder) {
```
lorenzleutgeb: Comment on line 237 applies here as well.

src/main/java/at/ac/tuwien/kr/alpha/solver/ArrayAssignment.java
AntoniusW: What we need here is an MBT-assigned, ordinary atom. Unfortunately, the assignment does not know which atom id represents an ordinary atom, this knowledge is currently only in the grounder. Hence the `getAtom` and the test for `BasicAtom` to decide if the atom is an ordinary one.

lorenzleutgeb: I opened #132 to follow up on this.

AntoniusW:
I think the architectural problem originating in the `ArrayAssignment.getSomeMBTAssignedAtom` has improved now since it relies on the now common `AtomTranslator` and no longer requires the grounder. Please have a look again @lorenzleutgeb and @rtaupe.

src/main/java/at/ac/tuwien/kr/alpha/grounder/structure/PredicateDependencyGraph.java
```diff
@@ -0,0 +1,127 @@
...
+					}
+					predicateDependencyGraph.addEdges(fromPredicate, predicatesInBody);
+				}
+			} else if (ruleHead instanceof DisjunctiveHead) {
+				for (Atom disjunctiveHeadAtom : ((DisjunctiveHead) ruleHead).disjunctiveAtoms) {
+					Predicate fromPredicate = disjunctiveHeadAtom.getPredicate();
+					predicateDependencyGraph.addEdges(fromPredicate, predicatesInBody);
+				}
+			} else {
+				oops("Unknown rule head encountered.");
```
rtaupe: `oops` alone does not do anything, you have to `throw oops`.

src/main/java/at/ac/tuwien/kr/alpha/solver/DefaultSolver.java
```diff
@@ -224,15 +227,142 @@ private boolean learnBackjumpAddFromConflict(ConflictCause conflictCause) {
...
+		ProgramAnalyzingGrounder analyzingGrounder = (ProgramAnalyzingGrounder) grounder;
+		// Justify one MBT assigned atom.
+		Integer toJustify = ((ArrayAssignment) assignment).getSomeMBTAssignedAtom();
+		if (LOGGER.isDebugEnabled()) {
+			LOGGER.debug("Searching for justification of " + toJustify + " / " + atomTranslator.atomToString(atomOf(toJustify)));
+			LOGGER.debug("Assignment is (TRUE part only):" + translate(assignment.getTrueAssignments()));
+		}
+		Set<Literal> reasonsForUnjustified = analyzingGrounder.justifyAtom(toJustify, assignment);
+
+		// Turn the justification into a NoGood.
```
rtaupe: I propose to extract lines 250-256 into a new method.

src/main/java/at/ac/tuwien/kr/alpha/solver/DefaultSolver.java
```diff
@@ -67,9 +72,11 @@
 	private boolean initialize = true;
 	private int mbtAtFixpoint;
 	private int conflictsAfterClosing;
+	private final boolean disableJustifications;
+	private boolean disableJustificationAfterClosing = true;	// Keep disabled for now, case not fully worked out yet.
```
rtaupe: Please create an issue so we can track this.

src/main/java/at/ac/tuwien/kr/alpha/solver/DefaultSolver.java
```diff
@@ -224,15 +227,142 @@ private boolean learnBackjumpAddFromConflict(ConflictCause conflictCause) {
...
+				continue;
+			}
+			if (!(atom instanceof RuleAtom)) {
+				// Ignore atoms other than RuleAtom.
+				toJustifyIterator.remove();
+				continue;
+			}
+			// For RuleAtoms in toJustify the corresponding ground body contains BasicAtoms that have been assigned FALSE in the closing.
+			// First, translate RuleAtom back to NonGroundRule + Substitution.
+			String ruleId = (String) ((ConstantTerm)atom.getTerms().get(0)).getObject();
```
rtaupe: ConstantTerm is a raw type. References to generic type ConstantTerm<T> should be parameterized

src/test/java/at/ac/tuwien/kr/alpha/grounder/SubstitutionTest.java
```diff
@@ -79,6 +80,38 @@ public void unifyTermsFunctionTermBinding() throws Exception {
...
+		VariableTerm varY = VariableTerm.getInstance("Y");
+		VariableTerm varZ = VariableTerm.getInstance("Z");
+		Term constA = ConstantTerm.getInstance("a");
+		Substitution left = new Substitution();
+		left.put(varX, varY);
+		left.put(varZ, varY);
+		Substitution right = new Substitution();
+		right.put(varX, constA);
+		Substitution merged = Substitution.mergeIntoLeft(left, right);
+		assertEquals(merged.eval(varY), constA);
```
rtaupe: The first parameter of `assertEquals` is `expected`, the second one is `actual`.

src/main/java/at/ac/tuwien/kr/alpha/grounder/structure/AnalyzeUnjustified.java
```diff
@@ -0,0 +1,419 @@
...
+				Substitution merged = Substitution.mergeIntoLeft(substitution, sigma);
+				// Ignore inconsistent merges.
+				if (merged == null) {
+					continue;
+				}
+				vNp.add(merged);
+			}
+			log("Adapting N to N'. Original N is {}", vN);
+			log("Adapted N' is {}", vNp);
+			// Line 8.
```
rtaupe: Please correct or remove line numbers

src/main/java/at/ac/tuwien/kr/alpha/solver/DefaultSolver.java
AntoniusW: Done, issue #138.

src/main/java/at/ac/tuwien/kr/alpha/solver/DefaultSolver.java
AntoniusW: Well, since this is a cast, doing so will result in an unchecked cast warning from the compiler. Due to type erasure there is no way for the runtime environment to be sure that the object of the ConstantTerm is a String. As far as I know there are only two options: using a generic cast and suppressing the unchecked cast warning, or doing it like above. The latter option seems preferable to me.

src/main/java/at/ac/tuwien/kr/alpha/grounder/Substitution.java
```diff
@@ -203,13 +289,22 @@ private static Substitution specializeSubstitution(Substitution substitution, Te
...
+						System.out.println("Problem.");
+					}
+				}
 				return substitution;
 			}
 			// The variable is bound, check whether its result is exactly the specific term.
 			// Note: checking whether the bounded term is more general than the specific one would yield
 			//       wrong results, e.g.: p(X,X) and p(f(A),f(g(B))) are incomparable, but f(A) is more general than f(g(B)).
-			if (substitutedGeneralTerm != specificTerm) {
+			if (substitutedGeneralTerm == specificTerm) {
```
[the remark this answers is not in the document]
AntoniusW: Since our terms all use an `Interner` and should only be constructable by `getInstance(..)` methods, it is safe to assume that two terms are equal if-and-only-if they are the same object.

src/main/java/at/ac/tuwien/kr/alpha/solver/DefaultSolver.java
rtaupe: The generic cast does _not_ produce a warning:
`String ruleId = (String) ((ConstantTerm<?>)atom.getTerms().get(0)).getObject();`
This option seems preferable to me because Eclipse IDE does not produce any warning for it (in contrast to the option currently implemented).

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/BasicAtom.java
```diff
@@ -95,7 +93,17 @@ public BasicAtom substitute(Substitution substitution) {
 			.map(t -> t.substitute(substitution))
 			.collect(Collectors.toList()));
 	}
-	
+
+	@Override
+	public BasicAtom normalizeVariables(String prefix, int counterStartingValue) {
```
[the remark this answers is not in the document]
AntoniusW: Done.

src/main/java/at/ac/tuwien/kr/alpha/common/AtomTranslator.java
```diff
@@ -1,12 +1,68 @@
...
+	 * @return true if the ground atom is already associated an integer ID.
+	 */
+	boolean contains(Atom groundAtom);
+
+	/**
+	 * Returns a list of currently known but unassigned atoms.
+	 * @param assignment the current assignment.
+	 * @return a list of atoms not having assigned a truth value.
+	 */
+	List<Integer> getUnassignedAtoms(Assignment assignment);
```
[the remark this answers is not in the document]
AntoniusW: Indeed, this `getUnassignedAtoms` could be in a better place. Actually, after PR #137 it would be best to place this functionality in the new `TrailAssignment` there.

Note that the `AtomTranslator` is now the interface to the `AtomStore`. Maybe they should be called differently?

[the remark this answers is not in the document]
rtaupe: I agree that `getUnassignedAtoms` should be in `TrailAssignment`.
If `AtomStore` stays the only implementation of `AtomTranslator`, one possibility would be to name the former `AtomStoreImpl` and the latter `AtomStore` ...

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/Atom.java
```diff
@@ -82,6 +83,17 @@ default Literal toLiteral() {
 	 */
 	Literal toLiteral(boolean positive);
 
+	default Atom renameVariables(String newVariablePrefix) {
+		ArrayList<VariableTerm> occurringVariables = new ArrayList<>(getBindingVariables());
+		occurringVariables.addAll(getNonBindingVariables());
+		Substitution renamingSubstitution = new Substitution();
+		int counter = 0;
+		for (VariableTerm variable : occurringVariables) {
+			renamingSubstitution.put(variable, VariableTerm.getInstance(newVariablePrefix + counter++));
```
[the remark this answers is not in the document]
rtaupe: Just noting that this question is still open.

AntoniusW:
All raised issues should be addressed now. Please have a look again.

PR #132, opened 2018-05-09
Split AtomStore.add and make Assignment iterable

lorenzleutgeb:


src/main/java/at/ac/tuwien/kr/alpha/common/Assignment.java
```diff
@@ -11,7 +11,7 @@
 import static at.ac.tuwien.kr.alpha.solver.ThriceTruth.FALSE;
 import static at.ac.tuwien.kr.alpha.solver.ThriceTruth.TRUE;
 
-public interface Assignment {
+public interface Assignment<T extends Assignment.Entry> extends Iterable<T> {
```
rtaupe: What is the reason for making `Assignment` parametrizable?
Now we get warnings in every place where `Assignment` is used as a raw type, so I suggest to either take this back or to add a type argument wherever `Assignment` is referenced.

src/main/java/at/ac/tuwien/kr/alpha/grounder/NaiveGrounder.java
```diff
@@ -575,6 +576,16 @@ public boolean isFact(Atom atom) {
 		return analyzeUnjustified.analyze(atomToJustify, currentAssignment);
 	}
 
+	@Override
+	public <E extends Assignment.Entry> Integer getSomeMBTAssignedAtom(Assignment<E> assignment) {
```
rtaupe: Why does this return `Integer` and not `int`?

lorenzleutgeb: TBH I do not know. This method was moved from `ArrayAssignment` to here, the signature is by @AntoniusW

src/main/java/at/ac/tuwien/kr/alpha/common/Assignment.java
lorenzleutgeb: Since the implementation of `NaiveGrounder#getSomeMBTAssignedAtom` wants to iterate over the assignment, I thought that implementing `Iterable` would be reasonable. Iterating over an assignment seems quite a natural request. However, since `Assignment.Entry` is just an interface, we need generics to hint what the concrete implementation of `Assignment.Entry` will be.
I missed those warnings and added wildcards where the raw type was referenced.

What is a little ugly is that now we have `Set<Integer> Assignment#getTrueAssignments` in addition to `Assignment#iterator()`. Removing `#getTrueAssignments` in favor of `#iterator()` is not straight forward since `NaiveSolver` would then need some implementation of `Assignment`. The growing difference between `NaiveSolver` and `DefaultSolver` is indeed having a bad influence on our architecture. We might split them more or even drop `NaiveSolver` at some point.

AntoniusW: Actually, the reason it wants to iterate the assignment should be an implementation detail of the assignment. If there were a collection of pointers to MBT-assigned atoms, it would be much more economical to just iterate that and not the whole assigment. If the assignment would somehow know which atom ids represent internal atoms and which represent atoms over the domain of the input ASP program, this method would not be part of the grounder at all.

The problem with the justification is that in some sense it reverses our information flow: previously, the grounder was given assignments and deriving new ground instances; now the solver starts looking at rules itself and by that may also arrive at unseen ground instances. I (partially) solved this issue by moving the justification analysis into the grounder and handing it the current assignment. Maybe a different approach would work better? But I am not seeing it yet. Having the justification analysis in the solver created even more problems because of dependencies on grounder-internal knowledge.

AntoniusW:
I have the feeling, that the second part of this PR (make Assignment iterable) moves into the wrong direction for the moment. The underlying issue is to avoid `BasicAtom` leaking into `Assignment`, which probably cannot be avoided, since we want to get an mbt-assigned ordinary atom, which are represented by `BasicAtoms`. How to find such an atom should be `Assignment`-internal.

I think the better solution would be to make AtomStore a common class that is used by grounder and solver components (and shared between them). It has a similar functionality as the `AtomTranslator` and maybe the `AtomTranslator` should not be implemented by the `Grounder` but by the `AtomStore` then. This seems much clearer to me. @lorenzleutgeb  what do you think about this?

Regarding the first part of the PR, I have no objections.

AntoniusW:
Hi @lorenzleutgeb and @rtaupe, thank you for raising the issue with the `Assignment` and the weird grounder information flow. I tried to address the root cause in commit 40355c3 and moved the atom-int translation fully out of the grounder in the `AtomTranslator` interface that is now implemented by the `AtomStore`. The `AtomStore` is then shared between grounder and solver component and can directly be queried to get the "type" of an atom represented by an int. This makes the `Grounder` interface much cleaner and I think it improves our architecture overall.

Since 40355c3 also contains the split of `AtomStore.add`, I close this PR. If any issues were not correctly addressed, please raise them again at #130. Thank you for your input!

PR #146, opened 2018-09-12
Re-enable some unit tests that were disabled to save resources during CI

rtaupe:


lorenzleutgeb:
I triggered a re-build on Travis CI, which shows not only timeouts but also failed test in terms of invalid solutions. That's strange...

src/test/java/at/ac/tuwien/kr/alpha/solver/HanoiTowerTest.java
```diff
@@ -82,12 +90,16 @@ private void testHanoiTower(int instance) throws IOException {
 	}
 
 	private void testHanoiTower(String instance) throws IOException {
+		if ("naive".equals(solverName)) {
```
AntoniusW: The default way to run tests conditionally seems to be `org.junit.Assume.assumeTrue(..)` and this should be placed in the `@Test` method.

src/test/java/at/ac/tuwien/kr/alpha/solver/HanoiTowerTest.java
```diff
@@ -82,12 +90,13 @@ private void testHanoiTower(int instance) throws IOException {
 	}
 
 	private void testHanoiTower(String instance) throws IOException {
+		ignoreTestForNaiveSolver();
```
AntoniusW: Please move the conditional-ignore to the level of the unit test and not nested inside some (private) testing code, i.e., make the call to `ignoreTestForNaiveSolver();` at the beginning of `testSimple()`.

src/test/java/at/ac/tuwien/kr/alpha/solver/OmigaBenchmarksTest.java
```diff
@@ -74,19 +77,19 @@ public void testLocstrat_400() throws IOException {
...
 
-	@Test(timeout = 10000)
+	@Test(timeout = 15000)
 	@Ignore("disabled to save resources during CI")
 	public void testReach_4() throws IOException {
 		test("reach", "reach-4.txt");
 	}
 
 	private void test(String folder, String aspFileName) throws IOException {
+		ignoreTestForNaiveSolver();
```
AntoniusW: Please move the conditional-ignore to the level of the unit test and not nested inside some (private) testing code, i.e., make the call to `ignoreTestForNaiveSolver();` at the beginning of `testReach_1()` and all other tests that should not be run with the naive solver. Actually, since these are the Omiga benchmarks, the naive solver should be able to run them in somewhat acceptable time. Please check what tests need to be ignored and which work fine.

src/test/java/at/ac/tuwien/kr/alpha/solver/OmigaBenchmarksTest.java
```diff
@@ -74,19 +77,19 @@ public void testLocstrat_400() throws IOException {
...
 
-	@Test(timeout = 10000)
-	@Ignore("disabled to save resources during CI")
+	@Test(timeout = 15000)
 	public void testReach_1() throws IOException {
 		test("reach", "reach-1.txt");
 	}
 
-	@Test(timeout = 10000)
+	@Test(timeout = 15000)
```
AntoniusW: Why increase the timeout and at the same time ignore the test completely?

rtaupe: OK, I take it back

rtaupe:
@AntoniusW thank you for your comments, they have been addressed. Please continue your review.

PR #152, opened 2018-11-21
Immediately ground already-ground rules by default

### LINKED ISSUE #149: Immediately ground already-ground rules by default
Currently, `NaiveGrounder` only grounds rules (including constraints) whose positive body is satisfied by the current partial assignment, even if the rule already is ground in the input program.
To increase solving efficiency, I propose "grounding" (i.e. forwarding to the solver in the form of nogoods) such rules immediately by default. In a future version of Alpha, we may want to support controlling this behaviour via command-line arguments.

**Example 1:**
```
a :- not b.
b :- not a.
c :- b.
```
The third rule is currently not grounded on the empty partial assignment.

**Example 2:**
```
a :- not b.
b :- not a.
:- b.
```
The constraint is currently not grounded on the empty partial assignment.

rtaupe:
Immediately ground already-ground rules by default by

- not stopping grounding a rule if a positive body atom is encountered that is not yet true (currently only for fully ground rules, but configurable),
- substituting ground atoms by themselves in this case, and
- considering all body literals as starting literals for the grounder in case a rule is fully ground.

Closes #149

### DIFF
--- a/src/main/java/at/ac/tuwien/kr/alpha/common/Rule.java
+++ b/src/main/java/at/ac/tuwien/kr/alpha/common/Rule.java
@@ -147,7 +147,7 @@ public class Rule {
 	}
 
 	public boolean isGround() {
-		if (!head.isNormal()) {
+		if (head != null && !head.isNormal()) {
 			throw oops("Called isGround on non-normal rule");
 		}
 		if (!isConstraint() && !((DisjunctiveHead)head).isGround()) {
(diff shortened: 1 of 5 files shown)


rtaupe:
Changes as discussed in person today:

- increase stability of `NaiveGrounderTest` by avoiding strict comparisons of string representations of nogoods
- respect `stopBindingAtNonTruePositiveBody` parameter in `NaiveGrounder` for non-ground rules, but always turn already-ground rules into nogoods immediately

PR #163, opened 2019-01-31
Fix facts in justification, add early arithmetics evaluation

### LINKED ISSUE #159: Program unsatisfiable because arithmetic terms are not evaluated
A program that is actually satisfiable is reported to be unsatisfiable, apparently because arithmetic terms are not evaluated in time.

Can be reproduced with the KnightTourWithHoles problem.

Encoding: `encoding.asp` from [ASP Competition 2015](http://aspcomp2015.dibris.unige.it/KnightTourWithHoles.zip)

Instance: Example from [ASP Competition 2013](https://www.mat.unical.it/aspcomp2013/KnightTour):
```
size(9).
forbidden(1,9). forbidden(2,9). forbidden(3,9). forbidden(4,9). forbidden(5,9). forbidden(6,9). forbidden(7,9). forbidden(8,9). forbidden(9,9).
forbidden(9,1). forbidden(9,2). forbidden(9,3). forbidden(9,4). forbidden(9,5). forbidden(9,6). forbidden(9,7). forbidden(9,8). forbidden(9,9).
```

AntoniusW:
Fixes #159.

### DIFF
--- a/src/test/java/at/ac/tuwien/kr/alpha/common/terms/TestMinusTerm.java
+++ b/src/test/java/at/ac/tuwien/kr/alpha/common/terms/TestMinusTerm.java
@@ -41,14 +41,14 @@ public class TestMinusTerm {
 
 	@Test
 	public void testNormalizeVariablesNoVariable() {
-		MinusTerm m2 = MinusTerm.getInstance(ConstantTerm.getInstance(2));
+		Term m2 = MinusTerm.getInstance(ConstantTerm.getInstance(2));
 		assertEquals(m2, m2.normalizeVariables(renamePrefix, counter));
 	}
 
 	@Test
 	public void testNormalizeVariablesWithVariable() {
-		MinusTerm mX = MinusTerm.getInstance(VariableTerm.getInstance("X"));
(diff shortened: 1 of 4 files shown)


src/test/java/at/ac/tuwien/kr/alpha/common/terms/TestMinusTerm.java
```diff
@@ -41,14 +41,14 @@
 
 	@Test
 	public void testNormalizeVariablesNoVariable() {
-		MinusTerm m2 = MinusTerm.getInstance(ConstantTerm.getInstance(2));
+		Term m2 = MinusTerm.getInstance(ConstantTerm.getInstance(2));
```
rtaupe: I do not understand the rationale behind the changes in this file.

AntoniusW: Rationale: if an ArithmeticTerm or MinusTerm is ground `getInstance` now returns the corresponding ConstantTerm, hence `MinusTerm.getInstance(..)` now only returns `Term` instead of `MinusTerm`. In the case here it returns a `ConstantTerm`.

Issue #162, opened 2019-01-28
Alpha ignores second input file if -i is missing


rtaupe:
If a user accidentally calls `java -jar Alpha.jar -i program.asp instance.asp` instead of `java -jar Alpha.jar -i program.asp -i instance.asp`, the input from `instance.asp` is silently ignored.

AntoniusW:
Issue should be resolved now, thanks to #164. Please confirm whether this is the case @rtaupe. Thanks.

rtaupe:
Fixed by e51364e7a9ec7f1e4792236609c698b92e353a05. Thanks @madmike200590!

PR #164, opened 2019-02-05
Commandline Interface Refactoring (Issue #162)

madmike200590:
Changes:
- Reworked command line argument parsing in separate class
- Separated external predicate handling into extra class (ExternalAtoms)
- Restructure Interface for all program invocations to go through Alpha class rather than duplicating that logic in Main
- Fixed issue #162, all positional arguments not preceded by an option latter will throw an error
- Added functionality to simultaneously accept input from any number of files and strings, e.g. "java -jar Alpha.jar -i file1.asp -i file2.asp -str "someFact." -str "otherFact :- someFact." -i file3.asp"


src/main/java/at/ac/tuwien/kr/alpha/config/IOptionHandler.java
```diff
@@ -0,0 +1,11 @@
+package at.ac.tuwien.kr.alpha.config;
+
+import org.apache.commons.cli.Option;
+import org.apache.commons.cli.ParseException;
+
+@FunctionalInterface
+public interface IOptionHandler<T> {
```
lorenzleutgeb: We don't prefix interfaces with a capital I. Just `grep` through the repo. Just for the sake of consistency I advise against it. Once there's an interface with prefix in the codebase, the second might follow, and then we end up mixing... Sorry if you prefer to have the prefix...

src/main/java/at/ac/tuwien/kr/alpha/config/AlphaContext.java
```diff
@@ -0,0 +1,27 @@
+package at.ac.tuwien.kr.alpha.config;
+
+/**
+ * Wrapper type for AlphaConfig and InputConfig.
+ */
+public class AlphaContext {
```
lorenzleutgeb: I dislike that we end up with `AlphaContext` and `AlphaConfig`, which might be confusing. Also, why is it a context and not just another config? You might well have `AlphaConfig` containing `InputConfig` and `SystemConfig`. Or have `AlphaConfig` contain `InputConfig`?

src/main/java/at/ac/tuwien/kr/alpha/config/IOptionHandler.java
madmike200590: Personally rather prefer the variant with prefix - But on the other hand it's really no trouble removing it, so I guess I'm just gonna do that

src/main/java/at/ac/tuwien/kr/alpha/config/AlphaContext.java
madmike200590: The idea behind calling this AlphaContext was that it wraps the 2 configs and is the top-level "config-like" structure, similar to an ApplicationContext in Spring.
AlphaConfig shouldn't directly contain InputConfig - The idea is that if you were to use Alpha as an API, you'd have one Alpha instance (initialized with an AlphaConfig) that represents the system. Whenever you want to solve a program (from whatever source), you'd create an InputConfig and pass it to your Alpha instance. In short, the separation between the two configs makes sure Alpha doesn't have an internal state that depends on any specific program.
However, I like your naming suggestion, so I'll rename the current AlphaConfig to SystemConfig and AlphaContext to AlphaConfig

src/main/java/at/ac/tuwien/kr/alpha/config/IOptionHandler.java
lorenzleutgeb: Yeah, I guessed so. Naming is hard. It's just for the sake of consistency really, not to bother you.

src/main/java/at/ac/tuwien/kr/alpha/config/AlphaContext.java
lorenzleutgeb: Makes sense. Thank you for elaborating. At least we don't mix `Context` and `Config` now.

src/main/java/at/ac/tuwien/kr/alpha/Alpha.java
```diff
@@ -27,201 +27,114 @@
...
+		Program retVal = new Program();
+		if (!cfg.getFiles().isEmpty()) {
+			retVal.accumulate(this.readProgramFiles(cfg.isLiterate(), cfg.getPredicateMethods(), cfg.getFiles()));
 		}
-
-		if (method.getGenericReturnType().getTypeName().startsWith(PredicateInterpretation.EVALUATE_RETURN_TYPE_NAME_PREFIX)) {
-			this.predicateMethods.put(name, new BindingMethodPredicateInterpretation(method));
-			return;
+		if (!cfg.getAspStrings().isEmpty()) {
+			retVal.accumulate(this.readProgramString(StringUtils.join(cfg.getAspStrings(), "\n"), cfg.getPredicateMethods()));
```
rtaupe: Not sure if it really makes a difference here, but afaik it is usually better to use `System.lineSeparator()` instead of `"\n"`.

src/main/java/at/ac/tuwien/kr/alpha/Main.java
```diff
@@ -27,256 +27,76 @@
...
-	private static final String DEFAULT_STORE = "alphaRoaming";
-	private static final String OPT_SEED = "seed";
-	private static final String OPT_DEBUG_INTERNAL_CHECKS = "DebugEnableInternalChecks";
-	private static final String DEFAULT_BRANCHING_HEURISTIC = Heuristic.NAIVE.name();
 
-	private static final java.util.function.Predicate<Predicate> DEFAULT_FILTER = p -> true;
+	private static final Logger LOGGER = LoggerFactory.getLogger(Main.class);
 
-	private static CommandLine commandLine;
+	private static final String ALPHA_CALL_SYNTAX = "java -jar alpha-bundled.jar" + System.lineSeparator() + "java -jar alpha.jar";
```
rtaupe: Isn´t it the case that the artefacts are usually called `Alpha-bundled.jar` and `Alpha.jar`? I think this is because the Github project name starts with an upper-case A.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ComparisonAtom.java
```diff
@@ -37,8 +37,7 @@
 import java.util.stream.Collectors;
 
 /**
- * Represents a builtin atom according to the standard.
- * Copyright (c) 2017-2018, the Alpha Team.
+ * Represents a builtin atom according to the standard. Copyright (c) 2017-2018, the Alpha Team.
```
rtaupe: If we keep this copyright statement, it should stay in its own line for better readability. We could also drop it, however, since it is redundant to the one in line 2.

src/main/java/at/ac/tuwien/kr/alpha/config/CommandLineParser.java
```diff
@@ -0,0 +1,301 @@
+package at.ac.tuwien.kr.alpha.config;
```
rtaupe: Since lots of content in this file have been copied from Main.java, the license header from the original file should also be copied. (This may also be the case for other files.)

src/main/java/at/ac/tuwien/kr/alpha/config/CommandLineParser.java
```diff
@@ -0,0 +1,301 @@
...
+import org.apache.commons.cli.Options;
+import org.apache.commons.cli.ParseException;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import at.ac.tuwien.kr.alpha.common.Predicate;
+
+/**
+ * Parses given argument lists (as passed when Alpha is called from command line) into <code>AlphaConfig</code>s and <code>InputConfig</code>s.
```
rtaupe: Please use `{@link AlphaConfig}` instead of `<code>AlphaConfig</code>` and `{@link InputConfig}` instead of `<code>InputConfig</code>`.

src/main/java/at/ac/tuwien/kr/alpha/grounder/NaiveGrounder.java
```diff
@@ -52,8 +52,7 @@
 import static java.util.Collections.singletonList;
 
 /**
- * A semi-naive grounder.
- * Copyright (c) 2016-2018, the Alpha Team.
+ * A semi-naive grounder. Copyright (c) 2016-2018, the Alpha Team.
```
rtaupe: If we keep this copyright statement, it should stay in its own line for better readability. We could also drop it, however, since it is redundant to the one in line 2.

src/main/java/at/ac/tuwien/kr/alpha/grounder/NaiveGrounder.java
```diff
@@ -484,7 +483,8 @@ public int register(NoGood noGood) {
 				throw oops("Grounded atom should be ground but is not");
 			}
 
-			if (factsFromProgram.get(substitutedAtom.getPredicate()) == null || !factsFromProgram.get(substitutedAtom.getPredicate()).contains(new Instance(substitutedAtom.getTerms()))) {
+			if (factsFromProgram.get(substitutedAtom.getPredicate()) == null
```
rtaupe: The many formatting changes in this file (and other files) make it hard to see whether anything besides formatting has changed at all. Please format only files (or, even better, lines) where you actually change semantics.

src/main/java/at/ac/tuwien/kr/alpha/Main.java
lorenzleutgeb: Depends on the name of the Gradle project inside of which the task to produce the bundled JAR is run. On my machine the project source code, i.e. this repository is placed inside a directory called "alpha" and not "Alpha" for personal (and therefore, subjective) convenience. This results in lowercase JAR names.

rtaupe: I realize that the directory name can be changed, that´s why I used the word "usually". No magic string that we are using here will always be right, but isn´t it the case that if you clone the repo from GitHub it will by default have an upper-case A?

rtaupe: We could also try to use the [actual name of the JAR file](https://stackoverflow.com/questions/14189162/get-name-of-running-jar-or-exe).

madmike200590: I did a quick test and the "actual name of jar file" variant works if I do it like this:
`String appname = new File(Main.class.getProtectionDomain().getCodeSource().getLocation().toURI()).getName();
		CommandLineParser commandLineParser = new CommandLineParser(appname, (msg) -> Main.exitWithMessage(msg, 0));`

That works fine both when I do a "java -jar" on the bundled jar and when I call the Main class directly with "java -cp Alpha-bundled.jar at.ac.tuwien.kr.alpha.Main".
**However**, you don't get any useful results from that if you don't have the class in a jar file when calling it from command line (I tested this using the eclipse "execute button" since I didn't want to type out the complete classpath).
Furthermore,  "...getLocation().toUri()" throws a URISyntaxException if whatever code location we use isn't representable as a URI - I'm not sure how likely that happening is, but I'm kinda reluctant to deal with the potential case for something that's really just a "nice to have".

The cleanest way I see to do this would be:
- adapt gradle config  to add the build artifact name into the META-INF/MANIFEST.MF file of the jar and read it from there
- if there is no manifest file (explicit check, no exception handler), fall back to "<your-alpha-jar>.jar"

However, even this doesn't address the (admittedly super-picky) point that we wouldn't be able to automatically test the format of the usage message with a unit test since the output would be influenced by how the test runner loaded the class.

Please let me know your thoughts!

src/main/java/at/ac/tuwien/kr/alpha/Alpha.java
madmike200590: changed

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ComparisonAtom.java
madmike200590: removed the lower copyright statement

src/main/java/at/ac/tuwien/kr/alpha/config/CommandLineParser.java
madmike200590: done

src/main/java/at/ac/tuwien/kr/alpha/config/CommandLineParser.java
madmike200590: done

src/main/java/at/ac/tuwien/kr/alpha/grounder/NaiveGrounder.java
madmike200590: done (see item below)

src/main/java/at/ac/tuwien/kr/alpha/grounder/NaiveGrounder.java
madmike200590: Sorry for all the formatting fails! I adapted my formatted config a few weeks ago, new commits should be better. For this file, I just simply reverted to the current version from master, this should take care of it.

src/main/java/at/ac/tuwien/kr/alpha/Main.java
rtaupe: To me this sounds fine. Thanks!

PR #169, opened 2019-03-04
Fix error in performance logs

### LINKED ISSUE #166: Performance logs: time mistakenly reset when answer set is found
After an answer set has been found, the time passed but not the number of decisions used to compute performance logs is reset. This means that the overall number of decisions is divided by the time that has passed since the last answer set has been found, which does not make sense.

rtaupe:
Performance logs (e.g. "Overall performance: 6350 decisions in 27.02s or 235.0111 decisions per sec. Overall replayed assignments: 0") now continue counting when an answer set is found, i.e. they measure performance across the whole solving process.

Fixes #166

### DIFF
--- a/src/main/java/at/ac/tuwien/kr/alpha/solver/DefaultSolver.java
+++ b/src/main/java/at/ac/tuwien/kr/alpha/solver/DefaultSolver.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2016-2018, the Alpha Team.
+ * Copyright (c) 2016-2019, the Alpha Team.
  * All rights reserved.
  *
  * Additional changes made by Siemens.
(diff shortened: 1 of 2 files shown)


src/main/java/at/ac/tuwien/kr/alpha/solver/DefaultSolver.java
```diff
@@ -88,17 +90,16 @@ public DefaultSolver(AtomStore atomStore, Grounder grounder, NoGoodStore store,
 				BranchingHeuristicFactory.getInstance(branchingHeuristic, grounder, assignment, choiceManager, random),
 				new NaiveHeuristic(choiceManager));
 		this.disableJustifications = disableJustifications;
+		this.performanceLog = new PerformanceLog(choiceManager, (TrailAssignment) assignment, 1000l);
```
AntoniusW: Why is `msBetweenOutputs` initialized with `10001` here?

rtaupe: It is not initialized with `10001` but with `1000l`, which preserves the behaviour we had before (one log output every 1000 ms). Admittedly, the two numbers are hard to distinguish visually. For some reason I had thought I needed to provide a `long` explicitly, which is not the case. The `l` has now been dropped.

rtaupe:
> 
> 
> I very much like this fix/refactoring and would suggest to add some additional command line parameters to control the logging. I think logging should be off by default and only enabled upon request.

Good idea. Since this has nothing to do with the fix at hand, however, I propose to handle it in a different issue.

Issue #173, opened 2019-03-14
Improve ReplayHeuristic


rtaupe:
`ReplayHeuristic` has been introduced for #172 and has been implemented in a minimally viable way.
In this issue, we collect improvements to be implemented in the future:

- provide replay choices as ground atoms (e.g. `holds(a)`) instead of signed solver-internal atom IDs
- provide replay choices via files instead of command-line arguments
- ...

PR #174, opened 2019-03-14
Replay heuristic

### LINKED ISSUE #172: Replay heuristic
Alpha shall provide a _replay heuristic_ that is fed with a list of predetermined choices (i.e. signed atoms) which it then chooses without modification.

If a choice specified by the list is not admissible or does not involve an active choice point, an exception shall be thrown.

rtaupe:
Introduces a _replay heuristic_ that is fed with a list of predetermined choices (i.e. signed atoms) which it then chooses without modification.

Fixes #172.

~~This pull request depends on #157, so #157 should be merged first.~~

Note that this is a minimum viable implementation for #172. Additional improvements are outsourced to #173.


lorenzleutgeb:
@rtaupe I marked this PR as "blocked" since (as you explicitly stated), #157 should land first.

rtaupe:
@AntoniusW thank you for your comments, they have been addressed. Please continue your review.

PR #176, opened 2019-03-26
Update JDKs in .travis.yml

lorenzleutgeb:

PR #178, opened 2019-03-26
Upgrade Gradle to 5.3 on Travis CI and AppVeyor

lorenzleutgeb:


build.gradle
```diff
@@ -99,14 +99,14 @@ jacocoTestReport {
 	//       AntlrTasks above). It does not make sense to include them in our coverage
 	//       report.
 	afterEvaluate {
-		classDirectories = files(classDirectories.files.collect {
+		getClassDirectories().setFrom(files(classDirectories.files.collect {
```
lorenzleutgeb: This change is to fix:

```
gradle build --warning-mode=all

> Configure project :
The JacocoReportBase.setClassDirectories(FileCollection) method has been deprecated. This is scheduled to be removed in Gradle 6.0. Use getClassDirectories().from(...)
        at build_933ksadryj435o3112hzqz5a1$_run_closure9$_closure17.doCall(/home/lorenz/src/github.com/alpha-asp/alpha/build.gradle:102)
        (Run with --stacktrace to get the full stack trace of this deprecation warning.)
```

AntoniusW:
Maybe this is the wrong place, but I recently stumbled across gradlew, the Gradle wrapper again and I am wondering now why we do not use this instead of the local Gradle version.
Using the Gradle wrapper would no longer require everyone to install Gradle locally for building Alpha and we would be sure that it always builds with the same Gradle version.
What are the arguments against it?

lorenzleutgeb:
> Using the Gradle wrapper would no longer require everyone to install Gradle locally for building Alpha and we would be sure that it always builds with the same Gradle version.

Correct.

> What are the arguments against it?

The downside that I see is that since the Gradle Wrapper is implemented in Java (`./gradlew` runs `java -jar gradle/wrapper/gradle-wrapper.jar`), which requires a JAR being checked in to the Git repository. That's something I'd like to avoid.

AntoniusW:
I lean towards using the Gradle Wrapper, because:
1) we do not push binary releases often, thus the only way to get a recent version of the Alpha is to build it directly from master.
2) Gradle is not installed in most standard OS distributions, making it another hurdle for someone just trying Alpha (or for anyone having to switch work environments -- like I did in the last year).

Since I would like to remove as many hurdles as possible, the cost of having a tiny .jar in the repo seem negligible to me.

lorenzleutgeb:
See #179, which IMO we should deal with after #178 and #176 are merged.

lorenzleutgeb:
> The AppVeyor build fails

I fixed this by also installing Gradle 5.3 on AppVeyor.

PR #179, opened 2019-03-26
Switch to Gradle Wrapper

lorenzleutgeb:


AntoniusW:
Travis fails for JDK 12 because it cannot be downloaded using the BCL license. AppVeyor fails, probably because it wants to execute gradle natively.

Can you fix those two?

lorenzleutgeb:
> Can you fix those two?

Done

PR #180, opened 2019-03-28
Enable Gradle Build Scans

lorenzleutgeb:
See https://scans.gradle.com/

With this addition we will have the possibility to access and share detailed information about the build process of Alpha, such as duration of build phases, an overview of dependencies, etc. To get an idea, check out a build scan from this PR's source branch: https://scans.gradle.com/s/j7kqgasr5a7am

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -7,10 +7,10 @@ before_build:
   - ps: Get-Command java
 
 build_script:
-  - gradlew assemble --no-daemon
+  - gradlew build --scan --stacktrace
 
 test_script:
-  - gradlew check --no-daemon
+  - gradlew check
 
 after_test:
   - ps: |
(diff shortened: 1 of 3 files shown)


AntoniusW:
This is a nice idea, but I have some concerns regarding privacy, since some information about the local machine gets published.  We should be sure that this automated publishing only occurs on the CI servers (or if explicitly requested).
As far as I understand, the build now checks whether the environment variable `CI` is defined and then always publishes. Since `CI` also enables the extensive test suite, this is not the best flag for making sure that the build actually runs on the CI server. Is there some way we can be sure to not run on anyones personal computer?

lorenzleutgeb:
Sure. I will remove the trigger via the environment variable `CI` and instead add the explicit command line flag `--scan` to `.travis.yml`. This way, the possibility that someone sets the environment variable for another purpose and then ends up having information uploaded is completely eliminated.

AntoniusW:
Seems reasonable now.
Are there any other opinions from the team? @rtaupe @madmike200590 ?

rtaupe:
> 
> 
> Seems reasonable now.
> Are there any other opinions from the team? @rtaupe @madmike200590 ?

Sounds good!

Issue #75, opened 2017-10-30
Optimization: Evaluate input program partially.


AntoniusW:
It is often the case that rules could fire by facts of the input program alone. For example, the following ASP program contains facts for the edges of a graph and a rule that makes all edges bi-directional, e.g., to consider edges to be undirected.
```
edge(1,2).
edge(2,4).
edge(5,4).
...
edge(X,Y) :- edge(Y,X).
...
```
Here, the rule could be fully evaluated by the grounder such that the solver has less work to do and several rounds of back-and-forth between grounder and solver can be skipped.

This kind of optimization allows dramatic speed-ups for many programs and is implemented by all state-of-the-art ASP solvers. This evaluation can be done using stratified negation or well-founded semantics, where the latter is more general but also more involved.

PR #185, opened 2019-04-11
Refactor program transforms

madmike200590:
###### Program refactoring
- `Program`: renamed to `AbstractProgram<R extends AbstractRule<? extends Head>>` and made immutable (immutability doesn't apply to InlineDirectives)
- new type `InputProgram extends AbstractProgram<BasicRule>`: the most general type of program, returned by the parser, rules can contain any constructs allowed by the parser
- new type `NormalProgram extends AbstractProgram<NormalRule>`: A program that has only rules with normal heads (`NormalHead` introduced as new Head subtype)
- new type `InternalProgram extends AbstractProgram<InternalRule>`: A program that is completely transformed and ready to be passed to the grounder and solver (replaces `ProgramAnalysis`, allows no aggregate literals and only normal rule heads)

Above changes were made in order to properly reflect "transformation stages" while performing program transformations and to have a compile-time guarantee that the grounder/solver only ever see programs that underwent the necessary preprocessing.
Programs are now (mostly) immutable, thereby making debugging easier - before/after comparisons for program transformations are quite easy now.
Note that Atoms, Literals and InlineDirectives should also be made immutable to properly guarantee immutability of programs.

###### Rule refactoring
- `Rule` renamed to `AbstractRule<H extends Head>` and made immutable (immutability does not apply to Literals)
- new type `BasicRule extends AbstractRule<Head>`: rules as they are returned by the parser, any types of heads possible, any kind of literals in body
- new type `NormalRule extends AbstractRule<NormalHead>`: rules that have a normal head
- `NonGroundRule` renamed to `InternalRule extends NormalRule`: rules that are ready to be passed to the grounder

The changes to rules reflect the abovementioned changes to programs and serve the same purpose

###### Refinements in program transformations
- made `ProgramTransformation`abstract and added type params, signature is now `ProgramTransformation<I extends AbstractProgram<?>, O extends AbstractProgram<?>` in order to properly reflect input- and output-contracts for individual transformations
- adapted transformations to no longer transform a (now immutable) program in place, but return new program instances
- changed `IntervalTermToIntervalAtom` and `VariableEqualityRemoval` to only work on `NormalProgram`s

###### Other changes
- added a method `Atom setTerms(List<Term>)` to `Atom` that creates a new Atom with the updated term list and implemented on suitable candidates (others throw UnsupportedOperationException) in order to make IntervalTermToIntervalAtom transformation sideeffect-free
- Commandline-Parameter "-dG": writes dependency graph of given program as .dot file (debug feature & prep work for partial rule evaulation, issue #75 )
- component graph implementation in place, but not (yet) used apart from unit tests, will be needed for issue #75


.gitignore
```diff
@@ -144,3 +144,6 @@ gradle-app.setting
 /junit.log
 .idea/modules/*.iml
 *.tokens
+gradle
```
lorenzleutgeb: No, I disagree. The Gradle Wrapper and the contents of `gradle` should be checked in.

See:
 * https://github.com/github/gitignore/blob/master/Gradle.gitignore
 * https://stackoverflow.com/questions/20348451

madmike200590: Fixed!

src/main/java/at/ac/tuwien/kr/alpha/Alpha.java
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2017-2019, the Alpha Team.
+ * Copyright (c) 2017-2018, the Alpha Team.
```
AntoniusW: We should keep that at 2019. :-)

madmike200590: I went through the code now and tried to fix up all the copyright notices I could find, should be better now

AntoniusW:
For some reason, a lot of the moved files are appearing as deleted and added again. As far as I know, git can track moved files (if done using git mv). Is it possible to do that? Reviewing is really cumbersome without it.

madmike200590:
> For some reason, a lot of the moved files are appearing as deleted and added again. As far as I know, git can track moved files (if done using git mv). Is it possible to do that? Reviewing is really cumbersome without it.

Unfortunately git doesn't alsways seem to get that a file was moved when I do it through Eclipse's "refactor move". To be honest, I wasn't aware of git mv, will look into it and see if I can tell git that those files were moved, but not sure if that is possible after the fact.

AntoniusW:
@madmike200590 you should also add your name/email to the `AUTHORS` and `CONTRIBUTORS` files.

lorenzleutgeb:
Regarding

> As far as I know, git can track moved files (if done using git mv). Is it possible to do that?

and

> Unfortunately git doesn't alsways seem to get that a file was moved when I do it through Eclipse's "refactor move".

Git does detect moves automatically, no matter whether `git mv` was used or not. If the change is commited, Git will make the attempt to associate renamed files. For the case where the content of files does not change, the ID of the underlying Git object will remain the same, so these renames are trivial to detect. However if some lines change, then Git resorts to some heuristic (don't ask me how it works n detail...).
To alleviate the problem, and making it a bit easier to follow through changes, separate the changes into at least two commits: In one of the commits only move files, in the other only change their content. Of course I know that this is slightly impractical, since often times the content of files mentions file names (package name, class name, imports, ...), so this gets tricky. But it is wise to split plain renaming and actual refactoring in terms of commits.

Consider these two questions on SO:
 - https://stackoverflow.com/questions/433111
 - https://stackoverflow.com/questions/7938582

The control knobs you have for detecting moves/copies more aggressively are:
 - [`git diff --find-renames[=<n>]`](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---find-renamesltngt) (try increasing `n`)
 - [`git diff --find-copies[=<n>]`](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---find-copiesltngt) (try increasing `n`)
 - [`git diff --find-copies-harder`](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---find-copies-harder)

src/main/java/at/ac/tuwien/kr/alpha/common/program/impl/InputProgram.java
```diff
@@ -0,0 +1,137 @@
...
+		public Builder accumulate(InputProgram prog) {
+			return this.addRules(prog.getRules()).addFacts(prog.getFacts()).addInlineDirectives(prog.getInlineDirectives());
+		}
+
+		public InputProgram build() {
+			return new InputProgram(this.rules, this.facts, this.inlineDirectives);
+		}
+	}
+
+	public boolean containsAggregates() {
```
AntoniusW: Apparently this method is not used (any longer) and can be deleted.

src/main/java/at/ac/tuwien/kr/alpha/common/program/impl/InputProgram.java
```diff
@@ -0,0 +1,137 @@
...
+ */
+public class InputProgram extends AbstractProgram<BasicRule> {
+
+	public static final InputProgram EMPTY = new InputProgram(Collections.emptyList(), Collections.emptyList(), new InlineDirectives());
+
+	public InputProgram(List<BasicRule> rules, List<Atom> facts, InlineDirectives inlineDirectives) {
+		super(rules, facts, inlineDirectives);
+	}
+
+	public InputProgram() {
```
AntoniusW: Apparently the default constructor is not used and would return a duplicate instance of `InputProgram.EMPTY`. I think it can be removed.

src/main/java/at/ac/tuwien/kr/alpha/grounder/transformation/impl/ChoiceHeadToNormal.java
```diff
@@ -25,34 +25,42 @@
  * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
-package at.ac.tuwien.kr.alpha.grounder.transformation;
+package at.ac.tuwien.kr.alpha.grounder.transformation.impl;
```
AntoniusW: Just for completeness: the copyright line above `Copyright (c) 2017-2018, the Alpha Team.` should be to `2019`.

src/main/java/at/ac/tuwien/kr/alpha/grounder/transformation/impl/ChoiceHeadToNormal.java
```diff
@@ -25,34 +25,42 @@
...
 import at.ac.tuwien.kr.alpha.common.terms.Term;
-
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Iterator;
-import java.util.List;
+import at.ac.tuwien.kr.alpha.grounder.transformation.ProgramTransformation;
 
 /**
  * Copyright (c) 2017-2018, the Alpha Team.
```
AntoniusW: `2019` again.

AUTHORS
```diff
@@ -15,4 +15,5 @@
 
 Antonius Weinzierl <antonius.weinzierl@gmail.com>
 Lorenz Leutgeb <lorenz.leutgeb@gmail.com> <lorenz.leutgeb@student.tuwien.ac.at> <e1127842@student.tuwien.ac.at>
-Siemens AG
\ No newline at end of file
+Michael Langowski <e1426581@student.tuwien.ac.at>
```
AntoniusW: Welcome onboard! :-)

src/main/java/at/ac/tuwien/kr/alpha/grounder/transformation/impl/ChoiceHeadToNormal.java
madmike200590: done

src/main/java/at/ac/tuwien/kr/alpha/common/program/impl/InputProgram.java
madmike200590: done

src/main/java/at/ac/tuwien/kr/alpha/common/program/impl/InputProgram.java
madmike200590: done

src/main/java/at/ac/tuwien/kr/alpha/grounder/transformation/impl/ChoiceHeadToNormal.java
madmike200590: done

rtaupe:
@madmike200590, are you still requesting my review for this pull request? If yes, please merge master into your branch. If no, please prefix the pull request's title by `WIP: `.

madmike200590:
@rtaupe thanks for reminding me, I should close that one since it's obsolete.
There's going to be a new PR in a few days once I'm fully done on #75 which will also contain the changes from this PR.

Issue #181, opened 2019-04-02
Revise domain-independent heuristics


rtaupe:
A while ago, a set of domain-independent branching heuristics (`BerkMin`, `DependencyDrivenHeuristic`, and their subclasses) has been developed. In contrast to the newly developed `VSIDS`, their runtime performance is not very good and must be improved. Also, some design decisions might have to be revised (e.g., do we need both methods `violatedNoGood` and `analyzedConflict`)?

rtaupe:
The TODO in `GeneralizedDependencyDrivenHeuristic#recordAtomRelationships` must also be fixed (use `HeapOfActiveChoicePoints#recordAtomRelationships`, which does similar things)

AntoniusW:
Another point is the member variable `boolean[] initializedActivityScores` in `HeapOfActiveAtoms`, which is currently read but not written to, i.e., for every new nogood, its literals are initialized again and again.

I would assume that after line 167 there should be some `initializedActivityScores[atom] = true;` but am not fully sure. @rtaupe can you shed some light on this?

rtaupe:
> 
> 
> Another point is the member variable `boolean[] initializedActivityScores` in `HeapOfActiveAtoms`, which is currently read but not written to, i.e., for every new nogood, its literals are initialized again and again.
> 
> I would assume that after line 167 there should be some `initializedActivityScores[atom] = true;` but am not fully sure. @rtaupe can you shed some light on this?

@AntoniusW thanks for noticing this. You are very right!
Since this is quite important for VSIDS, I just opened a new PR #187 and blocked #184.

AntoniusW:
With #187 merged, my above comment has been successfully addressed.

PR #187, opened 2019-04-13
Fix initialization of heuristic scores

rtaupe:
Addresses https://github.com/alpha-asp/Alpha/issues/181#issuecomment-482563854


AntoniusW:
Thanks for the quick reaction. I've had another thought about this and now wonder whether another solution would be better.

1. Currently, each atom is initialized once, so it may happen that the score of an atom is rather low at first but later grounded nogoods would increase it significantly. Now, I think it would be better to recompute the activity for as long as the atom did not receive an activityIncrement by VSIDS.
2. I just realized that `normalizeActivityScores` pushes all activity scores way below `1.0`, so an atom that is initialized after a normalization took place, receives a much higher value than all the other atoms. I am not fully sure what the best procedure is in such a case, but I think its score should be normalized as well (i.e., divided by `NORMALIZATION_THRESHOLD`).

rtaupe:
@AntoniusW thank you for your valuable suggestions, both have been addressed.

PR #188, opened 2019-04-15
Ignore .idea

### LINKED ISSUE #153: Inconsistent idea project configuration
Opening the existing project configuration from `.idea` with IntelliJ IDEA shows some errors: Files `.idea/modules/alpha.iml`, `.idea/modules/alpha_test.iml`, and `.idea/modules/alpha_main.iml` do not exist.

So the project configuration in `.idea` is inconsistent and should either be made consistent or removed from the repository.

lorenzleutgeb:
Proposal on how to resolve #153.


rtaupe:
I propose to also add `.idea/vcs.xml` to version control, or to add it to `.gitignore`. Currently it is not ignored but constantly re-generated by IDEA, without it IDEA "forgets" that the project is version-controlled by Git.

On my system, the file has the following contents:
```
<?xml version="1.0" encoding="UTF-8"?>
<project version="4">
  <component name="VcsDirectoryMappings">
    <mapping directory="$PROJECT_DIR$" vcs="Git" />
  </component>
</project>
```

rtaupe:
Following @lorenzleutgeb's [link](https://github.com/alpha-asp/Alpha/issues/153#issuecomment-483213087) to [StackOverflow](https://stackoverflow.com/questions/36372571/intellij-idea-and-gradle-why-there-are-3-modules-per-sub-module) shows the naming pattern `_main` and `_test`. So maybe the naming pattern has changed in a recent IDEA release ... I am using version 2019.01.

AntoniusW:
I just opened a fresh git clone (of the idea branch) with IDEA 2019.1 and the naming pattern is also `.main` and `.test`. There were no immediate problems with (File->New->Project from Existing Sources and selecting `build.gradle`), I did not try other ways.
What troubles me a bit is that:

1. After opening/importing from the gradle project, several files in the .idea folder are deleted: `copyright/profiles_settings.xml`, `checkstyle-idea.xml`, `codeStyleSettings.xml`, `compiler.xml`, and `encodings.xml` while `misc.xml` is changed and `vcs.xml` is created. Building the source works without trouble, and running all tests also does mostly (see next item). 
 2. If I click in the project view on the topmost folder and right-click "Run All Tests"  then I get a "No junit.jar found" error. "Run All Tests" works fine if the src/test folder is selected, though. Running the gradle tasks works out-of-the-box.

AntoniusW:
Given our recent exchange (and some testing), it seems that one option is to simply remove the .idea folder completely (and maybe put some explanation in the Readme stating that opening the `build.gradle` works out-of-the-box for intellij). The `.idea` folder then probably should also be included in gitignore.

The other option would be to remove `copyright/profiles_settings.xml`, `checkstyle-idea.xml`, `codeStyleSettings.xml`, `compiler.xml`, `encodings.xml`, and add `vcs.xml` and set in `misc.xml` the jdk version to 11.

What are your preferences? @rtaupe @lorenzleutgeb @madmike200590

lorenzleutgeb:
Considering how fragile `.idea` appears, and how cumbersome it is to get an understanding of a good solution for sharing it, I also thought about proposing to ignore `.idea` altogether.

That could lead to some trouble in the future, if we start using/relying on IntelliJ extensively and make exotic configurations. We would lose the possibility to share run configurations (which we haven't done up to now) and database connection settings (which we haven't done and probably will never need). From today's standpoint I don't see myself doing crazy things with IntelliJ that I would want to share with you via the repository...

If we ignore `.idea` we might keep that in mind if we observe strange behaviour/bugs in the future.

.gitignore
```diff
@@ -1,58 +1,11 @@
...
+# the project individually, since checking in IntelliJ related
+# files leads to annoying recurring conflicts/changes.
+.idea
 *.iws
+*.iml
 
-## Plugin-specific files:
-
-# IntelliJ
-/out/
```
AntoniusW: Why is the out directory no longer ignored? During my tests, I think I saw it once created, but I am not sure what caused it.

lorenzleutgeb: Because I decided to remove the default set of ignores from gitignore.io. But sure, we can include `/out/` if you like ...

AntoniusW:
Looks fine for me, let's give it a try without specific IDE configs.

rtaupe:
There is one minor remaining problem: When I import the Gradle project to IDEA, building from IDEA fails because classes generated by antlr are not found. One solution is to first build with `gradlew`, then building from IDEA will also succeed. I think we can live with this.

Thank you for your changes, I think this looks fine now!

PR #197, opened 2019-09-11
Enable forgetting of learned nogoods.

AntoniusW:
This PR addresses multiple interrelated issues:

- Conflict-driven learning uses the trail now (propagation levels no longer necessary).
- Solver-internal nogoods have activities now (that are bumped during learning).
- A first strategy to delete learned nogoods is implemented and enabled.
- Refactored and cleared several components of the core solver (TrailAssignment, NoGoodStoreAlphaRoaming, etc).


src/main/java/at/ac/tuwien/kr/alpha/solver/TrailAssignment.java
```diff
@@ -198,7 +207,7 @@ private void removeLastDecisionLevel() {
 		int start = trailIndicesOfDecisionLevels.get(getDecisionLevel());
 		ListIterator<Integer> backtrackIterator = trail.listIterator(start);
 		while (backtrackIterator.hasNext()) {
-			Integer backtrackAtom = backtrackIterator.next();
+			int backtrackAtom = atomOf(backtrackIterator.next());
```
rtaupe: Please document somewhere in the JavaDocs that the trail is now a trail of literals, not of atoms, but that the interfaces (e.g. `getAssignmentsToProcess`) still handle atoms.

src/main/java/at/ac/tuwien/kr/alpha/solver/TrailAssignment.java
```diff
@@ -767,7 +762,7 @@ public int hashCode() {
 
 		@Override
 		public String toString() {
-			return atom + "=" + value.toString() + "(DL" + decisionLevel + ", PL" + propagationLevel + ")"
+			return atom + "=" + value.toString() + "(DL" + decisionLevel + ")"
 				+ (hasPreviousMBT() ? "MBT(DL" + getMBTDecisionLevel() + ", PL-1 )" : "");
```
rtaupe: Shouldn´t the `PL` be removed here as well?

src/main/java/at/ac/tuwien/kr/alpha/solver/learning/GroundConflictNoGoodLearner.java
```diff
@@ -110,174 +117,144 @@ public GroundConflictNoGoodLearner(Assignment assignment) {
...
-			Assignment.Entry atomAssignmentEntry = firstUIPPriorityQueue.poll();
-			NoGood otherContributingNoGood = atomAssignmentEntry.getImpliedByRespectingLowerMBT();
-			if (otherContributingNoGood == null) {
-				// Case b), the other assignment is a decision.
-				return new ConflictAnalysisResult(null, atomAssignmentEntry.getWeakDecisionLevel(), true, resolutionAtoms);
+	private int backjumpLevelRemovingConflict(Antecedent violatedNoGood) {
+		int highestDL = 0;
+		int[] reasonLiterals = violatedNoGood.getReasonLiterals();
+		for (int literal : reasonLiterals) {
+			int literalDL = assignment.getWeakDecisionLevel(atomOf(literal)); //Math.min(assignment.getWeakDecisionLevel(atomOf(literal)), ((TrailAssignment) assignment).getOutOfOrderDecisionLevel(atomOf(literal)));
```
rtaupe: Why is this commented-out code here?

src/main/java/at/ac/tuwien/kr/alpha/solver/learning/GroundConflictNoGoodLearner.java
```diff
@@ -75,33 +79,36 @@ public int computeConflictFreeBackjumpingLevel(NoGood violatedNoGood) {
...
-		public final boolean clearLastChoiceAfterBackjump;
 		public final Collection<Integer> resolutionAtoms;
+		public final int lbd;
 
 		private ConflictAnalysisResult() {
 			learnedNoGood = null;
 			backjumpLevel = -1;
-			clearLastChoiceAfterBackjump = false;
 			resolutionAtoms = null;
+			lbd = -1;
```
rtaupe: `-1` is a magic number used at several places, so it should be moved to a constant with appropriate JavaDoc

src/main/java/at/ac/tuwien/kr/alpha/solver/learning/GroundConflictNoGoodLearner.java
```diff
@@ -110,174 +117,144 @@ public GroundConflictNoGoodLearner(Assignment assignment) {
...
-				return repeatAnalysisIfNotAssigning(currentResolutionNoGood, resolutionAtoms);
-				//return new ConflictAnalysisResult(currentResolutionNoGood, computeBackjumpingDecisionLevel(currentResolutionNoGood), false, noGoodsResponsible);
-			} else if (currentResolutionNoGood.size() > 32) {
-				// Break if resolved NoGood becomes too large.
-				// Remove all current-dl elements from the resolution NoGood and add the last choice, then backjump like usual.
-				return new ConflictAnalysisResult(null, conflictDecisionLevel, true, resolutionAtoms);	// Flag unsatisfiable abused here.
-				/*if (getAssignmentEntryRespectingLowerMBT(lastGuessedAtom).getDecisionLevel() <= conflictDecisionLevel) {
-					// If lastGuessedAtom is not unassigned after backjump, use repeatAnalysisIfNotAssigning.
-					return repeatAnalysisIfNotAssigning(replaceAllFromConflictDecisionLevelWithGuess(currentResolutionNoGood, conflictDecisionLevel, lastGuessedAtom), noGoodsResponsible, lastGuessedAtom);
+	private String printReasons(Collection<Integer> reasons) {
```
rtaupe: This method doesn´t actually _print_ anything, so I propose renaming it to `reasonsToString`
```suggestion
	private String reasonsToString(Collection<Integer> reasons) {
```

src/main/java/at/ac/tuwien/kr/alpha/solver/learning/GroundConflictNoGoodLearner.java
```diff
@@ -110,174 +117,144 @@ public GroundConflictNoGoodLearner(Assignment assignment) {
...
+			sb.append(isPositive(reasonLiteral) ? "+" + atomOf(reasonLiteral) : "-" + atomOf(reasonLiteral));
+			sb.append("@");
+			sb.append(assignment.getWeakDecisionLevel(atomOf(reasonLiteral)));
+			sb.append(", ");
+		}
+		sb.append("}");
+		return sb.toString();
+	}
+
+	private String printReasons(int[] reasons) {
```
rtaupe: This method doesn´t actually _print_ anything, so I propose renaming it to `reasonsToString`
```suggestion
	private String reasonsToString(int[] reasons) {
```

src/main/java/at/ac/tuwien/kr/alpha/solver/learning/GroundConflictNoGoodLearner.java
```diff
@@ -110,174 +117,144 @@ public GroundConflictNoGoodLearner(Assignment assignment) {
...
+		return printReasons(IntStream.of(reasons != null ? reasons : new int[0]).boxed().collect(Collectors.toList()));
```
rtaupe:
```suggestion
		return reasonsToString(IntStream.of(reasons != null ? reasons : new int[0]).boxed().collect(Collectors.toList()));
```

src/main/java/at/ac/tuwien/kr/alpha/solver/learning/GroundConflictNoGoodLearner.java
```diff
@@ -110,174 +117,144 @@ public GroundConflictNoGoodLearner(Assignment assignment) {
...
+	}
+
+	private ConflictAnalysisResult analyzeTrailBased(Antecedent conflictReason) {
+		LOGGER.trace("Analyzing trail based.");
+		if (assignment.getDecisionLevel() == 0) {
+			return ConflictAnalysisResult.UNSAT;
+		}
+		int numLiteralsInConflictLevel = 0;
+		ArrayList<Integer> resolutionLiterals = new ArrayList<>();
```
rtaupe:
```suggestion
		List<Integer> resolutionLiterals = new ArrayList<>();
```

src/main/java/at/ac/tuwien/kr/alpha/solver/learning/GroundConflictNoGoodLearner.java
```diff
@@ -110,174 +117,144 @@ public GroundConflictNoGoodLearner(Assignment assignment) {
...
+		ArrayList<Integer> resolutionAtoms = new ArrayList<>();
```
rtaupe:
```suggestion
		List<Integer> resolutionAtoms = new ArrayList<>();
```

src/main/java/at/ac/tuwien/kr/alpha/solver/learning/GroundConflictNoGoodLearner.java
```diff
@@ -110,174 +117,144 @@ public GroundConflictNoGoodLearner(Assignment assignment) {
...
+		int currentDecisionLevel = assignment.getDecisionLevel();
+		HashSet<Integer> seenAtoms = new HashSet<>();		// NOTE: other solvers use a global array for seen atoms, this might be slightly faster (initial tests with local arrays showed no significant improvement).
```
rtaupe:
```suggestion
		Set<Integer> seenAtoms = new HashSet<>();		// NOTE: other solvers use a global array for seen atoms, this might be slightly faster (initial tests with local arrays showed no significant improvement).
```

src/main/java/at/ac/tuwien/kr/alpha/solver/learning/GroundConflictNoGoodLearner.java
```diff
@@ -110,174 +117,144 @@ public GroundConflictNoGoodLearner(Assignment assignment) {
...
+		HashSet<Integer> processedAtoms = new HashSet<>();	// Since trail contains 2 entries for MBT->TRUE assigned atoms, explicitly record which seen atoms have ben processed to avoid processing seen atoms twice.
```
rtaupe:
```suggestion
		Set<Integer> processedAtoms = new HashSet<>();	// Since trail contains 2 entries for MBT->TRUE assigned atoms, explicitly record which seen atoms have ben processed to avoid processing seen atoms twice.
```

src/main/java/at/ac/tuwien/kr/alpha/solver/learning/GroundConflictNoGoodLearner.java
```diff
@@ -110,174 +117,144 @@ public GroundConflictNoGoodLearner(Assignment assignment) {
...
+			if (!assignment.isAssigned(atomOf(literal))) {
+				throw oops("Atom is not assigned: " + atomOf(literal));
 			}
+			occurringDecisionLevels.add(assignment.getWeakDecisionLevel(atomOf(literal)));
 		}
-		return resolvedLiterals;
+		return occurringDecisionLevels.size();
 	}
 
 	public ResolutionSequence obtainResolutionSequence() {
```
rtaupe: This method is never called, we could remove it

src/main/java/at/ac/tuwien/kr/alpha/solver/LearnedNoGoodDeletion.java
```diff
@@ -0,0 +1,92 @@
...
+		int originalSize = learnedNoGoods.size();
+		if (originalSize == 0) {
+			return;
+		}
+		int toDeleteMax = originalSize / 2;
+		long activitySum = 0;
+		for (WatchedNoGood learnedNoGood : learnedNoGoods) {
+			activitySum += learnedNoGood.getActivity();
+		}
+		double avgActivity = activitySum / originalSize;
```
rtaupe: Since `activitySum` is `long`, `avgActivity` will always be integer.
```suggestion
		double avgActivity = (double) activitySum / originalSize;
```

src/main/java/at/ac/tuwien/kr/alpha/solver/DefaultSolver.java
```diff
@@ -135,10 +135,16 @@ protected boolean tryAdvance(Consumer<? super AnswerSet> action) {
 			ConflictCause conflictCause = store.propagate();
 			didChange |= store.didPropagate();
 			LOGGER.trace("Assignment after propagation is: {}", assignment);
+			if (conflictCause == null) {
+				// Run learned NoGood deletion strategy.
+				store.cleanupLearnedNoGoods();
+			}
 			if (conflictCause != null) {
```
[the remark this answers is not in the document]
AntoniusW: I think due to the introduction of `disableNoGoodDeletion` the `else` would be wrong and would trigger if `conflictCause==null` and `disableNoGoodDeletion==true`.

src/main/java/at/ac/tuwien/kr/alpha/solver/Antecedent.java
```diff
@@ -0,0 +1,45 @@
...
+	void decreaseActivity();
+
+	/**
+	 * Tests whether two Antecedent objects have the same reason literals (irrespective of their order).
+	 * Note that both Antecedents are assumed to contain no duplicate literals.
+	 * @param l left Antecedent.
+	 * @param r right Antecedent
+	 * @return true iff both Antecedents contain the same literals.
+	 */
+	static boolean antecedentsEquals(Antecedent l, Antecedent r) {
```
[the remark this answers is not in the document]
AntoniusW: Moved to tests.

src/main/java/at/ac/tuwien/kr/alpha/solver/learning/GroundConflictNoGoodLearner.java
AntoniusW: Removed.

src/main/java/at/ac/tuwien/kr/alpha/solver/TrailAssignment.java
AntoniusW: Done.

AntoniusW:
Issues raised in the review have been addressed. @rtaupe please continue your review.

src/main/java/at/ac/tuwien/kr/alpha/solver/Antecedent.java
```diff
@@ -0,0 +1,17 @@
...
+ * An interface to reasons of implications as used internally by the solver. This is a lightweight {@link at.ac.tuwien.kr.alpha.common.NoGood} that only
+ * provides an array of literals (in some order) and has an activity that may change.
+ *
+ * Copyright (c) 2019, the Alpha Team.
+ */
+public interface Antecedent {
+
+	int[] getReasonLiterals();
+
+	void bumpActivity();
```
lorenzleutgeb: Did you consider naming this `increaseActivity`? I am asking because it appears that `bumpActivity` interplays with `decreaseActivity` and I expect some kind of dualism. Of that's the case then increase vs. decrease might highlight this better.

src/main/java/at/ac/tuwien/kr/alpha/solver/LearnedNoGoodDeletion.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.Iterator;
+
+import static at.ac.tuwien.kr.alpha.common.Literals.atomOf;
+
+/**
+ * Realizes a learned NoGood deletion strategy based on LBD and activity of NoGoods.
+ *
+ * Copyright (c) 2019, the Alpha Team.
+ */
+class LearnedNoGoodDeletion {
```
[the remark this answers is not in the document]
AntoniusW: The deletion of nogoods follows the ideas well established by other solvers. These usually provide a multitude of deletion strategies and command-line arguments to modify relevant parameters of each strategy. For now, Alpha only provides one deletion strategy that is modeled like the default one for ASP programs in clasp. I do not know how much we could gain by adding further deletion strategies.

src/main/java/at/ac/tuwien/kr/alpha/solver/Antecedent.java
AntoniusW: The names are inspired by names used in other solvers. `bumpActivity` increases by a constant while `decreaseActivity` decreases by a factor. So they indeed are dual in some way, but also different.

AntoniusW:
With the bug fixed, tests passing, and some unit tests added, I think this PR is ready for another round of reviewing.

src/test/java/at/ac/tuwien/kr/alpha/solver/learning/GroundConflictNoGoodLearnerTest.java
```diff
@@ -19,6 +19,7 @@
 
 	private final WritableAssignment assignment;
 	private final NoGoodStore store;
+	AtomStore atomStore;
```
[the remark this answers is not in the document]
AntoniusW: I think it is initialized in the constructor, is it not?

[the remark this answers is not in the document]
rtaupe: No, in the constructor a local variable with the same name is initialized.

src/main/java/at/ac/tuwien/kr/alpha/solver/heuristics/ReplayHeuristic.java
```diff
@@ -70,6 +70,10 @@ public int chooseLiteral() {
 			return DEFAULT_CHOICE_LITERAL;
 		}
 		int replayChoiceSignedAtom = choicesIterator.next();
+		if (replayChoiceSignedAtom == 0 ) {
```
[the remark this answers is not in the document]
AntoniusW: For debugging it was helping a lot to replay a whole sequence of guesses that involved multiple backtracks. If you just add the sequence without signalling "no more choices" at the appropriate points, then the solver will try to make a choice for something that is not an active choice point.

This allowed me to replay a whole run of the solver with a certain heuristic and see that the same sequence of choices (and backtracks) also triggered the same bug. Thus it helps to distinguish a bug inside a heuristic from one that is merely exhibited by a certain sequence of choices.

src/main/java/at/ac/tuwien/kr/alpha/solver/heuristics/ReplayHeuristic.java
```diff
@@ -70,6 +70,10 @@ public int chooseLiteral() {
...
+			// Use 0 to signal no more choices.
+			return DEFAULT_CHOICE_LITERAL;
+		}
 		int replayChoiceAtom = Math.abs(replayChoiceSignedAtom);
 		int replayChoiceLiteral = Literals.atomToLiteral(replayChoiceAtom, replayChoiceSignedAtom > 0);
```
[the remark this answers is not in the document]
AntoniusW: I think that then the lines below might trigger an exception, because `DEFAULT_CHOICE_LITERAL` is not an active choice atom. Also, I like that now the value of the 0 from the command-line argument is independent from the actual value of `DEFAULT_CHOICE_LITERAL`.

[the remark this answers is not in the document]
rtaupe: I agree.

Issue #113, opened 2017-12-22
Extended statistics


rtaupe:
Currently, Alpha is able to log a small number of statistical values on the solving process (cf. #51, #99, #102).
To gather more data, it could be beneficial to log other values as well. **Which ones should that be?**

For example, clasp outputs something like the following:
> Models       : 1+
> Calls        : 1
> Time         : 28.071s (Solving: 27.05s 1st Model: 27.04s Unsat: 0.00s)
> CPU Time     : 27.191s
> 
> Choices      : 133724
> Conflicts    : 103856   (Analyzed: 103856)
> Restarts     : 331      (Average: 313.76 Last: 167)
> Model-Level  : 26.0
> Problems     : 1        (Average Length: 0.00 Splits: 0)
>
> Lemmas       : 103856   (Deleted: 77119)
>>   Binary     : 1655     (Ratio:   1.59%)
>>   Ternary    : 2541     (Ratio:   2.45%)
>>   Conflict   : 103856   (Average Length:   33.4 Ratio: 100.00%)
>>   Loop       : 0        (Average Length:    0.0 Ratio:   0.00%)
>>   Other      : 0        (Average Length:    0.0 Ratio:   0.00%)
>
> Backjumps    : 103856   (Average:  1.26 Max:  48 Sum: 130701)
>>   Executed   : 103856   (Average:  1.26 Max:  48 Sum: 130701 Ratio: 100.00%)
>>   Bounded    : 0        (Average:  0.00 Max:   0 Sum:      0 Ratio:   0.00%)
> 
> Rules        : 105814
> Atoms        : 36917
> Bodies       : 67239
> Equivalences : 37788    (Atom=Atom: 303 Body=Body: 1758 Other: 35727)
> Tight        : Yes
> Variables    : 22212    (Eliminated:    0 Frozen:    0)
> Constraints  : 85168    (Binary:  77.7% Ternary:  22.2% Other:   0.1%)

AntoniusW:
Those statistics are indeed interesting. In principle, we could add all those statistics, although several of them are meaningless at the moment (e.g., Alpha currently does not delete Lemmas, we have no tight-ness analysis, and no variable elimination).

peschue:
As alpha works very different from gringo+clasp, some statistics will make sense in a more extended version.

I imagine it is interesting to count rules/atoms/bodies up to the first answer set and then separately from the first to the second and so on ...

[...]

PR #212, opened 2019-12-12
Add capability to count atoms for statistics

rtaupe:
Adds the capability to count atoms (grouped by classes implementing `Atom`). The numbers are currently printed only in debug log messages, similarly as numbers of nogoods in 84d86b3437841cc2007bd09863e3aaa4bc3d8dac. In the future, all this information shall be included in the statistics (cf. #113).

### DIFF
--- a/src/main/java/at/ac/tuwien/kr/alpha/solver/DefaultSolver.java
+++ b/src/main/java/at/ac/tuwien/kr/alpha/solver/DefaultSolver.java
@@ -27,23 +27,47 @@
  */
 package at.ac.tuwien.kr.alpha.solver;
 
-import at.ac.tuwien.kr.alpha.common.*;
-import at.ac.tuwien.kr.alpha.common.atoms.*;
+import at.ac.tuwien.kr.alpha.common.AnswerSet;
+import at.ac.tuwien.kr.alpha.common.Assignment;
+import at.ac.tuwien.kr.alpha.common.AtomStore;
+import at.ac.tuwien.kr.alpha.common.NoGood;
+import at.ac.tuwien.kr.alpha.common.Rule;
+import at.ac.tuwien.kr.alpha.common.atoms.Atom;
+import at.ac.tuwien.kr.alpha.common.atoms.BasicAtom;
+import at.ac.tuwien.kr.alpha.common.atoms.ComparisonAtom;
+import at.ac.tuwien.kr.alpha.common.atoms.Literal;
 import at.ac.tuwien.kr.alpha.common.terms.ConstantTerm;
 import at.ac.tuwien.kr.alpha.config.SystemConfig;
-import at.ac.tuwien.kr.alpha.grounder.*;
+import at.ac.tuwien.kr.alpha.grounder.Grounder;
+import at.ac.tuwien.kr.alpha.grounder.NonGroundRule;
+import at.ac.tuwien.kr.alpha.grounder.ProgramAnalyzingGrounder;
+import at.ac.tuwien.kr.alpha.grounder.Substitution;
 import at.ac.tuwien.kr.alpha.grounder.atoms.RuleAtom;
-import at.ac.tuwien.kr.alpha.solver.heuristics.*;
+import at.ac.tuwien.kr.alpha.solver.heuristics.BranchingHeuristic;
+import at.ac.tuwien.kr.alpha.solver.heuristics.BranchingHeuristicFactory;
+import at.ac.tuwien.kr.alpha.solver.heuristics.ChainedBranchingHeuristics;
+import at.ac.tuwien.kr.alpha.solver.heuristics.HeuristicsConfiguration;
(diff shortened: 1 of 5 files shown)


src/test/java/at/ac/tuwien/kr/alpha/solver/AtomCounterTests.java
```diff
@@ -0,0 +1,140 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class AtomCounterTests {
+
+	private final AtomStore atomStore = new AtomStoreImpl();
```
AntoniusW: I see no resetting (`@Before ...`) here, so I assume that the `AtomStore` and respective `AtomCounter` is shared among all tests. I guess the tests only work now, because they use the exact same atoms. Please fix that by introducing a proper resetting of the `AtomStore` and `AtomCounter`.

src/test/java/at/ac/tuwien/kr/alpha/solver/AtomCounterTests.java
```diff
@@ -0,0 +1,140 @@
...
+
+	private void expectGetNumberOfAtoms(AtomCounter atomCounter, Class<? extends Atom> classOfAtoms, int expectedNumber) {
+		assertEquals("Unexpected number of " + classOfAtoms.getSimpleName() + "s", expectedNumber, atomCounter.getNumberOfAtoms(classOfAtoms));
+	}
+
+	private void expectGetStatsByType(AtomCounter atomCounter, Class<? extends Atom> classOfAtoms, int expectedNumber) {
+		assertTrue("Expected number of " + classOfAtoms.getSimpleName() + "s not contained in stats string", atomCounter.getStatsByType().contains(classOfAtoms.getSimpleName() + ": " + expectedNumber));
+	}
+
+	@at.ac.tuwien.kr.alpha.common.atoms.external.Predicate
```
AntoniusW: Is this actually used? I think checking of the external atoms is a good idea and should be included in the tests.

rtaupe: Oh, I forgot to delete this. How could external atoms be tested here? `AtomStoreImpl#putIfAbsent` does not work for them because `ExternalAtom#isGround` returns false.

src/test/java/at/ac/tuwien/kr/alpha/solver/AtomCounterTests.java
rtaupe: Thanks. Fixed!

src/test/java/at/ac/tuwien/kr/alpha/solver/AtomCounterTests.java
madmike200590: I took a quick look at that  - It seems to me that ```ExternalAtom```s never come into contact with an ```AtomStore``` anyway since they're ```FixedInterpretationLiterals``` (or rather the enclosing Literals are). That means externals are fully evaluated by the grounder (Lines 418-432 in ```NaiveGrounder.bindNextAtomInRule```) and are never shown to the solver, meaning testing them here isn't necessary.

src/main/java/at/ac/tuwien/kr/alpha/solver/DefaultSolver.java
```diff
@@ -28,16 +28,10 @@
 package at.ac.tuwien.kr.alpha.solver;
 
 import at.ac.tuwien.kr.alpha.common.*;
-import at.ac.tuwien.kr.alpha.common.atoms.Atom;
-import at.ac.tuwien.kr.alpha.common.atoms.BasicAtom;
-import at.ac.tuwien.kr.alpha.common.atoms.ComparisonAtom;
-import at.ac.tuwien.kr.alpha.common.atoms.Literal;
+import at.ac.tuwien.kr.alpha.common.atoms.*;
```
madmike200590: I suggest avoiding "star imports". These can seriously affect the code's robustness to future changes in case a type with a name that already exists in the namespace is added in one of the "star-imported" packages. This would then result in files that weren't even changed not compiling anymore. Detailed explanation [here](https://stackoverflow.com/questions/147454/why-using-a-wild-card-with-a-java-import-statement-bad)

rtaupe: Fixed.

src/test/java/at/ac/tuwien/kr/alpha/solver/AtomCounterTests.java
rtaupe: Thanks, I removed the unused code.

rtaupe:
@madmike200590 @AntoniusW could either of you please merge this pull request?

madmike200590:
> @madmike200590 @AntoniusW could either of you please merge this pull request?

@AntoniusW I could do the merge some time today - if that's ok with you?

PR #133, opened 2018-05-10
Support for domain-specific heuristics encoded as heuristic directives

rtaupe:
These changes implement support for domain-specific heuristics encoded declaratively as heuristic directives within the input program.

PR #213, opened 2019-12-12
WIP: Extended support for domain-specific heuristics

rtaupe:
This is continuation of our work in #133. Work in progress.


AntoniusW:
It looks like the parser files have been adapted or split in multiple files -- or maybe some old parsing code is present here. Can you elaborate a bit what happened to the parser and why? Many thanks!

rtaupe:
> It looks like the parser files have been adapted or split in multiple files -- or maybe some old parsing code is present here. Can you elaborate a bit what happened to the parser and why? Many thanks!

Before my changes, we had a grammar file `ASPCore2.g4` which, despite its name, contained more features than the core language. I propose to split this into `ASPCore2.g4` containing only core features and `AlphaASP.g4` containing extensions.

Issue #214, opened 2019-12-19
Replace deprecated configuration in gradle.build


AntoniusW:
Currently, Gradle warns that our `gradle.build` uses deprecated configurations which will fail from Gradle 7.0 onwards. The detailed warning is:
```
The compile configuration has been deprecated for dependency declaration. This will fail with an error in Gradle 7.0. Please use the implementation configuration instead.
        at build_ejzegsudma2rbd3xma4zhkl9i$_run_closure3.doCall(/home/as/src/alpha/build.gradle:40)
        (Run with --stacktrace to get the full stack trace of this deprecation warning.)
The testCompile configuration has been deprecated for dependency declaration. This will fail with an error in Gradle 7.0. Please use the testImplementation configuration instead.
        at build_ejzegsudma2rbd3xma4zhkl9i$_run_closure3.doCall(/home/as/src/alpha/build.gradle:48)
        (Run with --stacktrace to get the full stack trace of this deprecation warning.)
The archiveName property has been deprecated. This is scheduled to be removed in Gradle 7.0. Please use the archiveFileName property instead.
        at build_ejzegsudma2rbd3xma4zhkl9i$_run_closure8.doCall(/home/as/src/alpha/build.gradle:87)
        (Run with --stacktrace to get the full stack trace of this deprecation warning.)
```

madmike200590:
corresponding PR #216 merged

PR #216, opened 2020-01-13
Fix deprecation warnings for gradle 7

madmike200590:
Addresses issue #214 

* changed deprecated dependency config "compile" to "implementation"
* changed deprecated property "archiveName" to use "archiveFileName" for bundledJar task
* exclude all "LICENSE.txt" and "NOTICE.txt" files from bundled jar

Note that LICENSE and NOTICE files are excluded from bundled jar in order to avoid having to handle duplicate files since all apache commons dependencies have their own files of that name which would have to be properly handled (ideally merged) otherwise. The aforementioned "proper handling" would require use of the gradle shadow plugin, which unfortunately is not gradle 7 compliant itself (as of version 5.2.0)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,11 +23,11 @@ repositories {
 /* The following configuration directive is a work-around for a fault in the Gradle
  * ANTLR plugin. It would require both antlr4 and antlr4-runtime at compile time and
  * at run time, which unnecessarily bloats our JARs. Only antlr4-runtime is needed.
- * We therefore remove this extension of antlr dependenIterablecies being compile dependencies
+ * We therefore remove this extension of antlr dependencies being compile dependencies
  * and reintroduce them on our own.
  */
 configurations {
-	compile {
+	implementation {
 		extendsFrom = extendsFrom.findAll { it != configurations.antlr }
 	}
 }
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -81,10 +81,13 @@ task bundledJar(type: Jar) {
...
 
 	from {
-		configurations.compile.collect { it.isDirectory() ? it : zipTree(it) }
+		configurations.compileClasspath.collect { it.isDirectory() ? it : zipTree(it) }
 	}
 
-	archiveName = "${project.name}-bundled.jar"
+	archiveFileName = "${project.name}-bundled.jar"
+
+	exclude('META-INF/LICENSE.txt')
```
lorenzleutgeb: I believe that these changes are in violation of the Apache License (v2, Section 4, Pragraphs a and d). If we remove licenses and notices from our distribution, then we should take care to reproduce them otherwise, e.g. print them as a special command or so.

madmike200590: I agree - however, the original state was in violation of that as well since both the license and notice files were overwritten with each other.
I'd rather not introduce a special command for something like that, especially since it would mean having those files not managed through gradle's dependency management anymore.

I think if we want to  fix this, the way to go would be using the [Shadow Plugin](https://github.com/johnrengelman/shadow). However, the plugin itself still produces deprecation warnings as of release 5.2.0, see [this issue](https://github.com/johnrengelman/shadow/issues/530).
Since this is not really a time-critical change, I'd suggest keeping this PR open until version 6.0 of the Shadow Plugin is released

AntoniusW: I would prefer a solution that does not add another dependency to the project. Having a low number of dependencies has a several of advantages, one of which is not having to wait for someone else to fix their (and now our) problems. Is there no way in Gradle to rename a file if a file with the same name already exists?

madmike200590: Since the build file is essentially a groovy script, that's doable.. I pushed a possible fix without using any other plugins, just numbering license and notice files. Feels a bit hacky tbh, but works..

AntoniusW: Given the status quo, I think it is an okay solution.

Issue #211, opened 2019-12-11
Alpha mistakenly proves Graph Colouring instance to be UNSAT on phase_saving_and_restarts branch


rtaupe:
On the `phase_saving_and_restarts` branch with phase saving and restarts enabled, the [new Graph Colouring encoding](https://github.com/alpha-asp/Alpha/files/3949801/encoding-new.asp.txt) is mistakenly proven to be UNSAT, e.g., on [instance 001](https://github.com/alpha-asp/Alpha/files/3949806/0001-graph_colouring-125-0.asp.txt). The [old encoding](https://github.com/alpha-asp/Alpha/files/3949803/encoding-alpha.asp.txt) works fine.

Command line:
`java -jar Alpha-bundled.jar -i encoding-new.asp -i instances/0001-graph_colouring-125-0.asp -n 1 -rs -b VSIDS_PHASE_SAVING`

PR #220, opened 2020-02-10
Fix bug in justification analysis, address issue #211.

AntoniusW:
Justification analysis used erroneously specific atom to test whether a given substitution is excluded by a more general one. Now a more general atom is constructed for the test. This fixes issue #211, this fix/PR is independent of the phase saving and restarts branch as the bug was just unearthed there but is present in master, too.


AntoniusW:
Since I did not find the time to construct a minimal working example / failing unit test for this issue in the last months and we have 100% coverage anyway, I'd like to suggest merging this PR sometime soon-ish (@rtaupe).

PR #226, opened 2020-02-21
Add additional simple support nogoods

rtaupe:
If there are multiple ground rules in the input program (i.e., before grounding!), each with a unique ground head, we can generate completion nogoods for each of them. So far this did not happen if the unique ground heads were of the same predicate.


rtaupe:
Build fails because of #239, but I do not think the cause for this issue comes from this pull request.

AntoniusW:
Thank you for the reminder. This PR adds certain completion nogoods, which is a good step in the right direction. There is, however, also a more principled approach in the direction of adding completion noogods currently under development (see PR #283 ). What is added here, also gets generated in #283 and more. I am currently working on the completion branch/PR (it needs a complete restructuring itself to cover more cases), with the goal of getting it ready in the next month. It is quite likely that this contribution here will be overwritten/redone/removed with #283. Do you still want it merged at this time?

rtaupe:
Yes please, because having ready features merged to master reduces maintenance efforts when we need them in other branches.

Thanks for your work on #283 though!

rtaupe:
I just found out that this pull request is buggy, because it does not check for multiple ground rules whose heads are of the same predicate if the heads are actually _different_.

For example, consider the following program:
```
h :- not a.
h :- not b.
```
With this pull request, completion nogoods for both rules are created individually, which is wrong.

Therefore, @AntoniusW I´m retracting this pull request and hope to be able to use your feature soon! I suggest you incorporate a unit test for this very condition in your branch.

PR #227, opened 2020-02-24
Deprecate old domain-independent heuristics

rtaupe:
Deprecate old domain-independent heuristics, and run solver tests only for non-deprecated heuristics.
The reason is that these old heuristics have not been maintained for a while and often perform poorly (#181). Sometimes they make CI break because tests take too much time using those heuristics.

### DIFF
--- a/src/main/java/at/ac/tuwien/kr/alpha/solver/heuristics/BranchingHeuristicFactory.java
+++ b/src/main/java/at/ac/tuwien/kr/alpha/solver/heuristics/BranchingHeuristicFactory.java
@@ -37,6 +37,11 @@ import java.util.stream.Collectors;
 
 public final class BranchingHeuristicFactory {
 
+	/**
+	 * The available domain-independent heuristics.
+	 * Some are deprecated because they perform poorly and have not been improved for some time,
+	 * however the code is kept for now so that it stays compatible when interfaces are refactored.
+	 */
 	public enum Heuristic {
 		NAIVE,
 		BERKMIN,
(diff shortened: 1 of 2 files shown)


src/main/java/at/ac/tuwien/kr/alpha/solver/heuristics/BranchingHeuristicFactory.java
```diff
@@ -41,19 +41,33 @@
 		NAIVE,
 		BERKMIN,
 		BERKMINLITERAL,
+		@Deprecated
```
madmike200590: I think it would be good to have a comment somewhere, explaining why those heuristics are deprecated. Maybe a short sentence in the class-level javadoc?

rtaupe: Good idea!

rtaupe:
@madmike200590 so far you have only approved this pull request. Could you please also merge it? See our [Contributing Guidelines](https://github.com/alpha-asp/Alpha/blob/master/CONTRIBUTING.md).

madmike200590:
> @madmike200590 so far you have only approved this pull request. Could you please also merge it? See our [Contributing Guidelines](https://github.com/alpha-asp/Alpha/blob/master/CONTRIBUTING.md).

Sorry, that one slipped through.. I wasn't aware I was supposed to merge as well when I did the review (wasn't sure if one review was enough).

PR #228, opened 2020-02-25
Provide a "standard library" of external atoms that is always available in programs

madmike200590:
### Provide a "standard library" of external atoms that is always available in programs
Adds a class `AspStandardLibrary` containing annotated predicate methods for frequently needed functions, enabling them to be used as external atoms in ASP code.

#### "Standard Library" predicate methods
* `stdlib_datetime_parse[DATETIME_STR, FORMAT_STR](Y, MO, D, H, MI, S)`
* `stdlib_datetime_is_before[DT1Y, DT1MO, DT1D, DT1H, DT1MI, DT1S, DT2Y, DT2MO, DT2D, DT2H, DT2MI, DT2S]`
* `stdlib_datetime_to_string[Y, MO, D, H, MI, S, FORMAT_STR](DT_STR)`
* `stdlib_string_matches_regex(STR, REGEX)`
* `stdlib_string_length[STR](LEN)`
* `stdlib_string_concat[S1, S2](S1S2)`

#### Java changes (summary)
* new base package `at.ac.tuwien.kr.alpha.api.externals` for classes mainly concerned with external atom support
* class `AspStandardLibrary` holding "standard lib" of external atoms
* modification to `InputConfig` making sure that the "standard lib" externals are loaded for every program
* ~~Utility class `ExternalUtils` for frequently needed convenience methods to convert from regular objects to sets of terms~~
* Bugfix in `Externals.scan(Package)`: This always scanned the complete classpath. Fixed so that only the given package is scanned
* New method `Externals.scan(Class)`: Only scan the given class for annotated methods

~~**NOTE**: The newly added test case `AspStandardLibraryTest.programWithStringStuff` is currently set to `@Ignore` due to a bug in Alpha's justification analysis (-> issue #229)~~

#### Bugfix: Issue #229

During testing of this PR, I found that justification analysis didn't work on `ExternalAtom`s since `ExternalAtom` did not implement `VariableNormalizableAtom`. Fixed this by implementing `VariableNormalizableAtom` on `ExternalAtom`. Since all implementations of this interface were using the exact same code, switched to a `default` implementation in the interface itself.


lorenzleutgeb:
I dislike the use of "_" to modularize the names. I'd argue that a module hierarchy like with Java (which uses "." for separation like DNS) would be better suited.

madmike200590:
> I dislike the use of "_" to modularize the names. I'd argue that a module hierarchy like with Java (which uses "." for separation like DNS) would be better suited.

The "strange" java method names came to be because I didn't realize until after I wrote it, that I could also pass a name to the `Predicate` annotation (I'll change that). As for the names themselves and how they're exposed on the ASP end, I'm also not really satisfied with the current state, but as long as (the Alpha variant of) ASP does not have a native concept of modules or namespaces, I'd like to keep the underscores. Assuming ASP language support for modules, I'd use "::" as a namespace separator, e.g. `stdlib::datetime::is_before`.

src/main/java/at/ac/tuwien/kr/alpha/api/externals/ExternalUtils.java
```diff
@@ -0,0 +1,67 @@
...
+
+	private ExternalUtils() {
+
+	}
+
+	/**
+	 * Convenience method for implementations of external atoms.
+	 * Returns a set containing the given value as it's only element.
+	 */
+	public static <T> Set<T> wrapAsSet(T value) {
```
lorenzleutgeb: I am surprised to see this. Wouldn't `java.util.Collections#singleton` work? I know that it returns an immutable set, but I would guess that the returned set isn't modified anyway.

src/main/java/at/ac/tuwien/kr/alpha/api/externals/stdlib/AspStandardLibrary.java
```diff
@@ -0,0 +1,157 @@
...
+	 * @param dt2Year   the year field for dt2
+	 * @param dt2Month  the month field for dt2
+	 * @param dt2Day    the day field for dt2
+	 * @param dt2Hour   the hour field for dt2
+	 * @param dt2Minute the minute field for dt2
+	 * @param dt2Second the second field for dt2
+	 * @return true if dt1 is before dt2 in time, false otherwise
+	 */
+	@Predicate
+	public static boolean stdlib_datetime_is_before(int dt1Year, int dt1Month, int dt1Day, int dt1Hour, int dt1Minute, int dt1Second,
```
lorenzleutgeb: I always had the idea in mind that calling Java methods on objects as external atoms should be hassle free. In this case, my preference would be to obtain two `LocalDateTime` objects an then call `isBefore` in some generic way, therefore eliminating the need to for a special "stdlib_datetime_is_before" implementation.

src/main/java/at/ac/tuwien/kr/alpha/common/terms/Terms.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Convenience methods for {@link Term}s.
+ * 
+ * Copyright (c) 2020, the Alpha Team.
+ */
+public final class Terms {
```
AntoniusW: I suggest to rename that class, since it is only utility methods.

src/test/java/at/ac/tuwien/kr/alpha/api/externals/stdlib/AspStandardLibraryTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+public class AspStandardLibraryTest {
+
+	//@formatter:off
+	private static final String STRINGSTUFF_ASP =
+			"string(\"bla\")."
+			+ "string(\"blubb\")."
+			+ "string(\"foo\")."
+			+ "string(\"bar\")."
+			+ ""
```
AntoniusW: This line literally adds nothing.

src/test/java/at/ac/tuwien/kr/alpha/api/externals/stdlib/AspStandardLibraryTest.java
```diff
@@ -0,0 +1,130 @@
...
+			+ "{ strcat(S1, S2) } :- string(S1), string(S2)."
+			+ "resultstring(SCAT) :- strcat(S1, S2), &stdlib_string_concat[S1, S2](SCAT)."
+			+ ""
```
AntoniusW: Again adding nothing here.

src/main/java/at/ac/tuwien/kr/alpha/api/externals/stdlib/AspStandardLibrary.java
```diff
@@ -0,0 +1,157 @@
...
+ * side-effect free.
+ * 
+ * Copyright (c) 2020, the Alpha Team.
+ */
+public final class AspStandardLibrary {
+
+	private AspStandardLibrary() {
+
+	}
+
```
AntoniusW: We should consider adding standard list operations (construct/head/tail/member/append/length) also to the library.

AntoniusW:
> ... Assuming ASP language support for modules, I'd use "::" as a namespace separator, e.g. `stdlib::datetime::is_before`, since dot is already being used as a rule terminator

Since we have a quite powerful parser, it should be no problem to use "." as namespace separator (we already do it for intervals). I would not want to resort to C/C++ parsing workarounds.

src/main/java/at/ac/tuwien/kr/alpha/api/externals/ExternalUtils.java
madmike200590: yep, you're right, I simply didn't think of that. Fixed!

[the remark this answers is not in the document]
madmike200590: Because I didn't think of `Collections.singleton` ;-) Fixed!

src/main/java/at/ac/tuwien/kr/alpha/api/externals/ExternalUtils.java
```diff
@@ -0,0 +1,67 @@
...
+	 * Convenience method for implementation of external atoms.
+	 * Takes a single {@link ConstantTerm} and wraps it into a
+	 * Set<List<ConstantTerm>>
+	 * 
+	 * @param <T>  the value type of the term
+	 * @param term the term to wrap
+	 * @return a Set<List<ConstantTerm<T>>> that contains the input term as it's
+	 *         only value
+	 */
+	public static <T extends Comparable<T>> Set<List<ConstantTerm<T>>> wrapSingleTerm(ConstantTerm<T> term) {
```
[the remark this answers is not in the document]
madmike200590: Fixed!

src/main/java/at/ac/tuwien/kr/alpha/api/externals/stdlib/AspStandardLibrary.java
madmike200590: In principle, I agree. However, I don't see that working in Alpha in the immediate future...
I think in a perfect world, I'd be able to have `LocalDateTime` as a "type" in ASP on which the "<"-operator is defined and mapped to the java implementation of `LocalDateTime#isBefore`.

Without any kind of ASP type system, I'd rather not directly call methods on objects from ASP, it just seems to be inviting chaos (side-effects, symmetry of comparison operations, etc..). While all that can also happen with my version from above, writing a dedicated method at least forces people to think about what a method is doing before exposing it to the ASP solver.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/VariableNormalizableAtom.java
```diff
@@ -1,15 +1,31 @@
...
+	 * @param prefix               the variable prefix V in front of the counter.
 	 * @param counterStartingValue the initial value i of the counter.
-	 * @return the Atom where all variables are renamed and enumerated left-to-right.
+	 * @return the Atom where all variables are renamed and enumerated
+	 *         left-to-right.
 	 */
-	Atom normalizeVariables(String prefix, int counterStartingValue);
+	default Atom normalizeVariables(String prefix, int counterStartingValue) {
+		List<Term> renamedTerms = Term.renameTerms(this.getTerms(), prefix, counterStartingValue);
+		return new BasicAtom(this.getPredicate(), renamedTerms);
```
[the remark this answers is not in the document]
madmike200590: I didn't think of that, thanks for pointing it out! I'll try to have a fix today

src/main/java/at/ac/tuwien/kr/alpha/common/terms/Terms.java
madmike200590: Quoting Joshua Bloch, author of "Effective Java" and one of the JVM devs, [on utility classes](https://www.informit.com/articles/article.aspx?p=1216151&seqNum=4) :
> [...] They can also be used to group static methods, including factory methods (Item 1), for objects that implement a particular interface, in the manner of java.util.Collections. [...]
So for a utility class providing helpers around a specific type, it's actually pretty much a convention to use the type name in plural, and I personally think it makes for easily readable code.

src/main/java/at/ac/tuwien/kr/alpha/common/terms/Terms.java
```diff
@@ -0,0 +1,35 @@
+package at.ac.tuwien.kr.alpha.common.terms;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Convenience methods for {@link Term}s.
```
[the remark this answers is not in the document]
madmike200590: done

src/main/java/at/ac/tuwien/kr/alpha/common/terms/Terms.java
```diff
@@ -0,0 +1,35 @@
...
+	public static <T extends Comparable<T>> List<ConstantTerm<T>> asTermList(T... values) {
+		List<ConstantTerm<T>> retVal = new ArrayList<>();
+		for (T value : values) {
+			retVal.add(ConstantTerm.getInstance(value));
+		}
+		return retVal;
+	}
+
+	@SafeVarargs
+	public static <T extends Comparable<T>> List<ConstantTerm<String>> asSymbolicTermList(T... values) {
```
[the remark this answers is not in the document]
madmike200590: nope, I thought I'd need it somewhere, but didn't. Removed it.

src/main/java/at/ac/tuwien/kr/alpha/common/terms/Terms.java
AntoniusW: That is a good argument. :-)
Following the linked text, I'd also suggest to make the private constructor throw an exception and add a comment that this should not be instantiated.

src/main/java/at/ac/tuwien/kr/alpha/config/InputConfig.java
```diff
@@ -17,13 +17,16 @@
 	public static final int DEFAULT_NUM_ANSWER_SETS = 0;
 	public static final boolean DEFAULT_WRITE_XLSX = false;
 	public static final String DEFAULT_OUTFILE_PATH = "alphaAnswerSet"; // current directory, files named "alphaAnswerSet.{num}.{ext}"
+	// standard library externals are always loaded
+	// public static final Map<String, PredicateInterpretation>
+	// DEFAULT_LOADED_EXTERNALS = Externals.getStandardLibraryExternals();
```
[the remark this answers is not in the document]
madmike200590: done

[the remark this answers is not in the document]
madmike200590: done

src/main/java/at/ac/tuwien/kr/alpha/common/terms/Terms.java
madmike200590: done

src/main/java/at/ac/tuwien/kr/alpha/config/InputConfig.java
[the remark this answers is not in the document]
madmike200590: > I guess that commentary should be removed.

done

src/test/java/at/ac/tuwien/kr/alpha/api/externals/stdlib/AspStandardLibraryTest.java
madmike200590: fixed

src/test/java/at/ac/tuwien/kr/alpha/api/externals/stdlib/AspStandardLibraryTest.java
madmike200590: fixed

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/VariableNormalizableAtom.java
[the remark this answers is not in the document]
madmike200590: fixed

src/main/java/at/ac/tuwien/kr/alpha/api/externals/stdlib/AspStandardLibrary.java
madmike200590: I'd be a fan of that. However, I think construct and append shouldn't be part of (at least the first iteration of) list support. Both operations could be understood as "write operations", therefore adding a procedural dimension to programs that's hard to handle with current ASP semantics.
Should we use Issue #76 to track ideas about that? Seems like the right place..

madmike200590:
> > ... Assuming ASP language support for modules, I'd use "::" as a namespace separator, e.g. `stdlib::datetime::is_before`, since dot is already being used as a rule terminator
> 
> Since we have a quite powerful parser, it should be no problem to use "." as namespace separator (we already do it for intervals). I would not want to resort to C/C++ parsing workarounds.

I was more thinking of readability than parsing. "." is also the logical candidate for struct membership (or whatever we want to call it), which I think would also be a nice ASP feature at some point. So having a separate operator for namespace membership would make for easier to read code. But like a lot of things, I guess this is also a matter of taste ;)

madmike200590:
Java methods for standard library externals are now renamed to conform to Java's camel-casing convention, (snake-cased) ASP names are provided as argument of the predicate annotation.

The bug in `normalizeVariables` that made every normalized atom a `BasicAtom` is also fixed.
@AntoniusW do you think there's anything still missing, or can this PR be merged?

src/main/java/at/ac/tuwien/kr/alpha/api/externals/stdlib/AspStandardLibrary.java
AntoniusW: Well, without construct I don't see a way to construct a list ... ?
By the way, in Prolog a list is just a special function term, so `[a,b,c]` is equal to `cons(a, cons(b, cons(c, nil)))`. However, if we implement lists via external atoms we have to be careful whenever lists are modified, because logically the modified list is a new object, i.e., in such cases the underlying Java list must be duplicated too.

AntoniusW:
> > > ... Assuming ASP language support for modules, I'd use "::" as a namespace separator, e.g. `stdlib::datetime::is_before`, since dot is already being used as a rule terminator
> > 
> > 
> > Since we have a quite powerful parser, it should be no problem to use "." as namespace separator (we already do it for intervals). I would not want to resort to C/C++ parsing workarounds.
> 
> I was more thinking of readability than parsing. "." is also the logical candidate for struct membership (or whatever we want to call it), which I think would also be a nice ASP feature at some point. So having a separate operator for namespace membership would make for easier to read code. But like a lot of things, I guess this is also a matter of taste ;)

That is one of the things, I think, Java did right compared to C++. The "." denotes that-thing-inside-the-other-thing and it does not matter if one is a namespage, a class, a method, or an object. Usually, it is clear from the context which kind it actually is, hence no need for the programmer to make it explicit. I like that. :-)

src/main/java/at/ac/tuwien/kr/alpha/api/externals/stdlib/AspStandardLibrary.java
```diff
@@ -0,0 +1,157 @@
...
+ * 
+ * All functions exposed by this class are guaranteed to be stateless and
+ * side-effect free.
+ * 
+ * Copyright (c) 2020, the Alpha Team.
+ */
+public final class AspStandardLibrary {
+
+	private AspStandardLibrary() {
+
```
AntoniusW: Please `throw new AssertionError(Terms.class.getSimpleName() + " is a non-instantiable utility class!");` here too.

src/main/java/at/ac/tuwien/kr/alpha/api/externals/stdlib/AspStandardLibrary.java
madmike200590: The problem I see with construction of lists via externals is this:
```
% forgive the syntax, just prototyping here...
a_list_as_fact(lst[1, 2, 3]). % no problem, list is a fact and will always exist

% Assume that LST is a (list-)function term, the rule reads content from SRC using 
% the external atom ext_content_list when some_other_condition(SRC) holds.
content_list(LST) :- content_src(SRC), &ext_content_list[SRC](LST), some_other_condition(SRC).
% This is a problem! Since the external atom is evaluated when the rule is grounded
% - but before it is clear whether it can actually fire - every failed attempt to ground this rule
% will create a new (Java-)list instance on the heap. 
```
The useless list instances mentioned should (I think, without having gone through the code with a fine-toothed comb..) be garbage-collected at some point, but I still see a potential memory leak here, or at least a dangerous (even if  temporary) heap-bloat.

src/main/java/at/ac/tuwien/kr/alpha/api/externals/stdlib/AspStandardLibrary.java
madmike200590: done

src/main/java/at/ac/tuwien/kr/alpha/api/externals/stdlib/AspStandardLibrary.java
AntoniusW: WIthout completely following this example, I still agree that native lists might come with intricate memory issues. Maybe we should discuss any further thoughts on that on #76 as you suggested.

PR #233, opened 2020-03-09
Improve grounder selectivity.

AntoniusW:
- IndexedInstanceStorage joins instances based on most selective
  attribute.


src/main/java/at/ac/tuwien/kr/alpha/grounder/IndexedInstanceStorage.java
```diff
@@ -166,23 +166,34 @@ public void removeInstance(Instance instance) {
...
+					mostSelectiveTermPosition = i;
+				}
 			}
 		}
-		return -1;
+		return mostSelectiveTermPosition;
 	}
 
-	public List<Instance> getInstancesFromPartiallyGroundAtom(Atom substitute) {
+	List<Instance> getInstancesFromPartiallyGroundAtom(Atom substitute) {
```
madmike200590: Is there a specific reason for this method to have default visibility? If so, I think there should be a comment explaining that since it's rather unusual..

madmike200590: As per our discussion, this was just a question out of curiosity... PR approved

PR #236, opened 2020-03-12
Avoiding Integer objects.

AntoniusW:
Alpha currently creates a lot of temporary `Integer` objects, most of it by autoboxing. This PR significantly reduces the amount of those (on some instance from 144 million down to 4 million temporary `Integer` objects).

The changes are as follows:

- Introduction of `IntIterator` instead of `Iterator<Integer>`.
- `TrailAssignment` stores the trail as `int[]` instead of `ArrayList<Integer>`.
- Logger calls that print `int` (which get autoboxed to `Integer`) moved inside corresponding if-statements.
- Additionally avoid creation of never-needed `BinaryAntecedent` by lazily postponing their instantiation, using a long-living object as placeholder (i.e., `ShallowAntecedent`).

### DIFF
--- a/src/main/java/at/ac/tuwien/kr/alpha/common/IntIterator.java
+++ b/src/main/java/at/ac/tuwien/kr/alpha/common/IntIterator.java
@@ -1,0 +1,20 @@
+package at.ac.tuwien.kr.alpha.common;
+
+/**
+ * An iterator returning raw int integers instead of Integer objects (for efficiency).
+ *
+ * Copyright (c) 2020, the Alpha Team.
+ */
+public interface IntIterator {
+
+	/**
+	 * @return true if the iterator has more elements.
+	 */
+	public abstract boolean hasNext();
+
+	/**
+	 * @return the next int in the iteration.
+	 */
+	public abstract int next();
+
+}
--- a/src/main/java/at/ac/tuwien/kr/alpha/common/IntIterator.java
+++ b/src/main/java/at/ac/tuwien/kr/alpha/common/IntIterator.java
@@ -10,11 +10,11 @@ public interface IntIterator {
 	/**
 	 * @return true if the iterator has more elements.
 	 */
-	public abstract boolean hasNext();
+	boolean hasNext();
 
 	/**
 	 * @return the next int in the iteration.
 	 */
-	public abstract int next();
+	int next();
 
 }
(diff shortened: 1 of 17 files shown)


src/main/java/at/ac/tuwien/kr/alpha/common/IntIterator.java
```diff
@@ -0,0 +1,20 @@
...
+ * An iterator returning raw int integers instead of Integer objects (for efficiency).
+ *
+ * Copyright (c) 2020, the Alpha Team.
+ */
+public interface IntIterator {
+
+	/**
+	 * @return true if the iterator has more elements.
+	 */
+	public abstract boolean hasNext();
```
madmike200590: Public interface methods should have no visibility modifier and no `abstract` since they are `public abstract`by default. See: https://stackoverflow.com/questions/5418464/methods-visibility-in-interface

src/main/java/at/ac/tuwien/kr/alpha/solver/NaiveSolver.java
```diff
@@ -277,8 +278,22 @@ private void doBacktrack() {
...
 
+	private static IntIterator fromIterator(Iterator<Integer> it) {
+		return new IntIterator() {
+			@Override
+			public boolean hasNext() {
+				return it.hasNext();
+			}
+
+			@Override
+			public int next() {
```
madmike200590: Not sure if this method makes sense.. `it.next()` gives an `Integer` (already boxed), and here it's unboxed again, so an `IntIterator` that is obtained through this method probably does not reduce autoboxing. It may be that the compiler is smart enough to see that the boxing/unboxing could be skipped, but I suspect that is not the case (thanks to type erasure). Hard to verify wiithout reading bytecode though..

src/main/java/at/ac/tuwien/kr/alpha/common/IntIterator.java
AntoniusW: Fixed.

src/main/java/at/ac/tuwien/kr/alpha/solver/NaiveSolver.java
AntoniusW: Your analysis is correct, this method does not avoid `Integer` objects. The surrounding anonymous class just wraps an `Iterator<Integer>` inside an `IntIterator` for the sole reason of using it inside the `NaiveSolver`, which is a totally non-optimized (and simple) version of a solver anyway.

Issue #239, opened 2020-03-17
Alpha mistakenly proves Hanoi instance to be UNSAT on add_simple_completion branch


rtaupe:
On the `add_simple_completion` branch, `HanoiTowerTest#testSimple` fails. The exact reason is unclear to me, but somehow the learnt nogoods are not processed correctly.

Notes:
  * seed has been fixed to 0
  * #220 does not help
  * I do not think this has something to do specifically what has been implemented on the `add_simple_completion` branch. Rather, this issue probably pops up because different nogoods are learnt.

These are all the nogoods that are learned:
  * `{ +3 +471 +475 -504 } = {+(succ(1, 2)), +(on(1, 5, 6)), +(on(1, 6, 7)), -(on(2, 5, 6))}` (by `analyzeTrailBased`)
  * `{ +1 +3 -95 +309 +459 +461 +463 } = {+(succ(0, 1)), +(succ(1, 2)), -(move(0, 6)), +(where(2, 5)), +(on(0, 5, 6)), +(on(0, 6, 7)), +(on(0, 7, 8))}` (by `analyzeTrailBased`)
  * `{ +1 +463 +465 -479 } = {+(succ(0, 1)), +(on(0, 7, 8)), +(on(0, 8, 9)), -(on(1, 7, 8))}` (by `analyzeTrailBased`)
  * `{ +1 +3 +135 -145 +254 +465 -648 } = {+(succ(0, 1)), +(succ(1, 2)), +(move(2, 8)), -(move(1, 9)), +(where(0, 2)), +(on(0, 8, 9)), -(on(1, 2, 9))}` (by `analyzeTrailBased`)
  * `{ +19 +34 +49 +64 +79 +94 +109 +124 +139 +235 } = {+(noMove(0, 1)), +(noMove(0, 2)), +(noMove(0, 3)), +(noMove(0, 4)), +(noMove(0, 5)), +(noMove(0, 6)), +(noMove(0, 7)), +(noMove(0, 8)), +(noMove(0, 9)), +(diskMoved(0))}` (by `justifyMbtAndBacktrack` and `noGoodFromJustificationReasons`)
  * `{ -34 } = {-(noMove(0, 2))}` (by `analyzeTrailBased`)

I have converted these nogoods to the following constraints:
```
:- succ(1, 2), on(1, 5, 6), on(1, 6, 7), not on(2, 5, 6).
[...]

PR #240, opened 2020-03-18
Count nogoods only if they are really added.

### LINKED ISSUE #238: NoGoods are counted multiple times if they are stored multiple times
It may happen that the same nogood is added to the `NoGoodStore` multiple times.
Examples:
 * In `HanoiTowerTest#testSimple` on the `add_simple_completion` branch, the nogood `{ +19 +34 +49 +64 +79 +94 +109 +124 +139 +235 } = {+(noMove(0, 1)), +(noMove(0, 2)), +(noMove(0, 3)), +(noMove(0, 4)), +(noMove(0, 5)), +(noMove(0, 6)), +(noMove(0, 7)), +(noMove(0, 8)), +(noMove(0, 9)), +(diskMoved(0))}` is added twice. The first time it is created by `DefaultSolver#justifyMbtAndBacktrack`, the second time from `DefaultSolver#noGoodFromJustificationReasons`.
  * In `DefaultSolver#addAndBackjumpIfNecessary`, there is a loop that (potentially) adds the same nogood again and again.

Since `NoGoodStoreAlphaRoaming#add` does not check for uniqueness of the nogood's ID (related issue: #87), the same nogood is then counted multiple times by `NoGoodCounter`.

One solution to this issue could be to invoke `NoGoodCounter#add` from `NogoodRegistry#register(NoGood)` instead of from `NoGoodStoreAlphaRoaming#add`.

Low prio because it only affects statistics.

rtaupe:
Fixes #238

### DIFF
--- a/src/main/java/at/ac/tuwien/kr/alpha/solver/NoGoodStoreAlphaRoaming.java
+++ b/src/main/java/at/ac/tuwien/kr/alpha/solver/NoGoodStoreAlphaRoaming.java
@@ -1,5 +1,5 @@
-/**
- * Copyright (c) 2016-2019, the Alpha Team.
+/*
+ * Copyright (c) 2016-2020, the Alpha Team.
  * All rights reserved.
  *
  * Additional changes made by Siemens.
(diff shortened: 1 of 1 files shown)


rtaupe:
Note: In the case of #239, the last learnt nogood, `{ -34 } = {-(noMove(0, 2))}`, is now not counted anymore (because it leads to a conflict).

Issue #237, opened 2020-03-16
Add support for unit tests in Alpha's input language


madmike200590:
Currently, unit-testing of ASP code requires one to write some kind of external script or application that triggers the solver and then validates the result.
It would greatly ease application development to have built-in support for unit-tests.

An idea how this could look in code would be:
```
% Example program that lists even and odd numbers up to a maximum value. 
% One answer set for evens and odds, respectively. 

1 {count_even; count_odd} 1.
num(0) :- count_even.
num(1) :- count_odd.

% "max_value(NUM)" needs to be passed as a separate program snippet
num(N) :- num(K), N = K + 2, N < MAX, max_value(MAX).

%%% UNIT TESTS %%%
@test(smoke_test_evens){ % this is test test case with name "smoke_test_evens"
    input { % test case input is an arbitrary list of facts (no rules!)
        max_value(10)
    }
    verify(expect: 2){ 
        % ASP program that is solved together with each answer set,
        % value for "expect" specifies the number of expected answer sets for the test case,
        % may be "unsat", any positive integer or a positive integer range

        :- num(11).
        :- num(12).
    }
}
```
A unit test T on a program P counts as failed iff the program `A union T` is unsatisfiable for any answer set A in AS(P).

To support the development workflow, one could then call Alpha with a commandline flag `--test` to run tests for a program. Tests would not be executed when the solver is called regularly for a program that includes tests.

Running included tests for ASP parts of a larger application could also be triggered through build systems like maven or gradle (e.g. trigger ASP tests as part of "test" phase via maven exec plugin)

madmike200590:
Another possible syntax style would be the following (thanks to @power-fungus for providing this!)
```
#test smoke_test_evens : {% clingo style pragmas
    max_value(10). % flat input facts

    % assertions as pragma
    assertion_error("num(11) incorrectly in answer set") :- num(11).
    assertion_error("num(12) incorrectly in answer set") :- num(12).
}
```

madmike200590:
Based on further discussions and the ideas above, I propose the following for a first version of this feature:

### Unit Test Syntax
* Test cases are identified by `#test NAME(expected: NUM_ANSWERS)`
* Unit test input and verification part are separated (as initially proposed)
* Input part must consist of **facts only**
* Verification part may contain arbitrary ASP code, but **must not have choices** (see "Test Evaluation" below)

#### EBNF-Draft
```
TEST_MARKER = "#test";
[...]

PR #243, opened 2020-04-01
WIP: Support for unit tests in ASP code (issue #237)

madmike200590:
**The branch is still a work in progress, creating this PR for transparency. Since this will be a rather visible feature with big impact, I believe it makes sense for everyone to be able to play around with it as early as possible**

## Add support for unit tests to Alpha's input language

Currently, unit-testing of ASP code requires one to write some kind of external script or application that triggers the solver and then validates the result.
It would greatly ease application development to have built-in support for unit-tests.

An idea how this could look in code would be:
```
% Example program that lists even and odd numbers up to a maximum value. 
% One answer set for evens and odds, respectively. 

1 {count_even; count_odd} 1.
num(0) :- count_even.
num(1) :- count_odd.

% "max_value(NUM)" needs to be passed as a separate program snippet
num(N) :- num(K), N = K + 2, N < MAX, max_value(MAX).

%%% UNIT TESTS %%%
#test smoke_test_evens(expected: 2) { % this is test test case with name "smoke_test_evens"
% value for "expect" specifies the number of expected answer sets for the test case,
% may be "unsat" or any positive integer
    input { % test case input is an arbitrary list of facts (no rules!)
        max_value(10).
    }
    verify { 
        % ASP program that is solved together with each answer set,

        :- num(11).
        :- num(12).
    }
}
```

### Unit Test Syntax
* Test cases are identified by `#test NAME(expected: NUM_ANSWERS)`
* Unit test input and verification part are separated (as initially proposed)
* Input part must consist of **facts only**
* Verification part may contain arbitrary ASP code, but **must not have choices** (see "Test Evaluation" below)

#### EBNF-Draft
```
TEST_MARKER = "#test";
TEST_ID = ('a'..'z') ( 'A'..'Z' | 'a'..'z' | '0'..'9' | '_' )*;  // same as regular ASP constant
EXPECT_ANSWERS = "unsat" | ([1..9][0..9]*);

unit_test = test_def '{' 
    [input]
    verify
    '}'
;

// "expected" keyword optional
test_def = TEST_MARKER ' ' TEST_ID '(['expected: '] EXPECTED_ANSWERS ')'; 
input = {fact};
verify = program; // arbitrary ASP program
```
### Test Evaluation

Tests are evaluated by Alpha as follows
```
TP = U union TI
TA = solve(TP)
if(not_matches(size(TA), TE)){
    fail(T)
}
for(AS : TA){
    TR = solve(P(AS) union TV)
    if(size(TR) != 1){ // for this to occur, `TV` must contain choices, which is not permitted
        fail(T)
    }
    if(assertionError(TR)){
        fail(T)
    }
}
pass(T)
```
Program components, intermediate results, and functions:
* Program `U`: The tested unit, i.e. the "regular" ASP code Alpha is called with
* Test `T`: One Unit-Test, consisting of Test-Input `TI` and Test-Verifier `TV`
* Program `TP`: The program resulting from adding the test input facts (`TI`) to the program under test `U`
* Set of Test Answer Sets `TA`: The answer sets of `TP`
* `size(ASS)`: returns the number of answer sets in the set `ASS`
* Size specifier `TE`: The expected value for the number of answer sets of `TP` (`unsat` or positive integer)
* `not_matches(size, expected)`: true iff the integer `size` does not conform to the expected answer sets specifier `expected`
* Program `P(AS)`: The program resulting from transforming the atoms in an answer set to a program consisting of facts only
* Test-Result Answer Set `TR`: The answer set of the program resulting from solving `P(AS) union TV`, i.e. the test verifier with added facts `P(AS)`
* `assertionError(AS)`: checks if an answer set `AS` contains assertion errors, i.e. `assertionError(AS)` is true iff `AS` contains atoms of any predicate with name `assertion_error` of any arity
* `fail(T)` lets the test `T` fail
* `pass(T)` lets the test `T` pass

Intuitively, the test input `TI` of a test `T` is added to the program under test `U` and solved. If the number of answer sets does not conform to the expectation of the test, the test fails. Every answer set of the resulting program is transformed into facts and combined with the test verifier `TV`. The answer sets resulting from this "verification program" are checked for assertion errors. If an assertion error occurs, the test fails. A test passes if and only if:
* The number of answer sets of `TP` conforms to what is expected by the test
* No answer set of the verification program contains assertion errors

#### Assertion Errors
Rules deriving assertion errors are created by **rewriting all constraints in `TV`** in order to make debugging of failed tests easier.
For example, the constraint `:- must_have(X, Y), not no_go(Y).` is rewritten to the rule
```
assertion_error(X, Y, ":- must_have(X, Y), not no_go(Y).") :- must_have(X, Y), not no_go(Y).
```
If the assertion specified by that rule fails, the error message displayed by Alpha is:
```
Assertion error in test TEST_NAME: Constraint ":- must_have(X, Y), not no_go(Y)" fails for ground instances {
    (X, Y) = (bla, blubb),
    (X, Y) = (foo, bar)
}
```
(assuming the assertion error occurs for ground substitutions `(bla, blubb)` and `(foo, bar)`)

#### Separation of "input" and "verify" parts of a test
Although a "flat" syntax would look cleaner, I'm heavily leaning towards the separated variant. Without this, test evaluation would have to be performed in one step (test program = `TestCode union programUnderTest`), thereby introducing the possibility of (unintentionally) influencing the number of answer sets of the program under test with verification code.
#### Scope of a test
Implementation-wise the program under test `U` is intended to be all the input Alpha is called with (potentially multiple files along with command-line-supplied strings), however, I propose the following guidelines for "good style when writing unit tests":
* Test code contained in the same file as implementation code **refers to that code only** and is not meant to be run together with additional code
* Test code targeted at implementation code that is spread out across multiple files should be written in a **dedicated test-suite file** that contains only unit tests (functionally, I'd call that an integration test then).
#### Commandline-interface for unit tests
By default, a commandline call like `java -jar Alpha.jar -i someFile.asp` will **not** run any unit tests contained in `someFile.asp`. The idea is that in most real-world uses there would be more than one input file which would make it necessary to disable tests using some flag most of the time.
To run unit tests on `someFile.asp`, the command would be `java -jar Alpha.jar --test -i someFile.asp`.


lorenzleutgeb:
I saw that notice at the very top of the PR and would like to make sure that you know about the following GitHub feature: https://github.blog/2019-02-14-introducing-draft-pull-requests/

madmike200590:
> I saw that notice at the very top of the PR and would like to make sure that you know about the following GitHub feature: https://github.blog/2019-02-14-introducing-draft-pull-requests/

Thanks for the heads-up! As I've proven with that notice, I didn't know that feature :sweat_smile: . I'll try converting this one to a draft, hope that's possible somehow..

AntoniusW:
This would be a very good addition to Alpha. Since the state where this draft-PR branched off the master is quite old, it might be easier to start with a fresh branch.

Regarding the parser grammar, here is a suggestion how the ANTLR grammer could be extended to parse the test directives:
```
directive : directive_enumeration | directive_test;  // NOT Core2 syntax, allows solver specific directives. Further directives shall be added here.

directive_test : SHARP 'test' ID PAREN_OPEN 'expected' COLON NUMBER PAREN_CLOSE CURLY_OPEN test_input test_verify+ CURLY_CLOSE;

test_input : 'input' CURLY_OPEN program CURLY_CLOSE;

test_verify : test_verify_forall | test_verify_exists;

test_verify_forall : 'verify' 'forall' CURLY_OPEN program CURLY_CLOSE;

test_verify_exists : 'verify' 'exists' CURLY_OPEN program CURLY_CLOSE;
```

There are two things to note:

1.  The `test_input` is allowed to contain any program, not just a collection of facts. That this part contains only facts needs to be checked separately (e.g. in the `ParseTreeVisitor` or even later).
2. This grammar will create a problem for IDs, since all special identifiers between ' and ' lead to implicit lexer rules that all have higher precedence than the `ID` rule. That means with this grammar it is no longer possible to have a fact like `test(foo).` or a constant `verify`. The solution is to turn the `ID` lexer rule into a parser rule and accept all the keywords occurring here: `id : ID | FORALL | EXISTS | TEST | EXPECTED | VERIFY;` and to replace all `ID` with `id` in the grammar, as well as the `'verify'` etc with `VERIFY` etc.

madmike200590:
Outdated - realized in PR #358

Issue #244, opened 2020-04-03
String-typed ConstantTerms for strings and constant symbols mistakenly equal


madmike200590:
It seems that the `symbolic` flag for `ConstantTerm<String>`s isn't taken into account when checking equality.
The following test illustrating this currently fails (see branch `bugfix_term_equality`):
```
@Test
public void testStringTermEquality() {
	ConstantTerm<String> stringConstant = ConstantTerm.getInstance("string");
	ConstantTerm<String> constantSymbol = ConstantTerm.getSymbolicInstance("string");
	Assert.assertNotEquals(stringConstant, constantSymbol);
}
```
The term called `stringConstant` represents the quoted string `"symbol"` in ASP code,
while the other term, `constantSymbol`, refers to the constant symbol `symbol` in code. These are not the same thing and should therefore not be equal.

madmike200590:
Further analysis showed that at least the "internalizing" still works correctly - `stringConstant` and `constantSymbol` are distinct instances in the term cache. This is due to the fact that the value of `symbolic` is taken into account by `ConstantTerm#hashCode`.

As a preliminary fix, I added the following check to `ConstantTerm#equals`:
```
if (this.symbolic != that.symbolic) {
	return false;
}
```

However, the above fix still leaves one further issue:
The assertion `Assert.assertNotEquals(0, stringConstant.compareTo(constantSymbol));` currently fails, meaning both terms are equal with respect to `compareTo`, which should not be the case

AntoniusW:
I guess this is a bug in `ConstantTerm.compareTo` where in line 120 we have
```
if (other.object.getClass() == this.object.getClass()) {
  return this.object.compareTo((T) other.object);
}
```
I guess what is missing there is `&& other.symbolic == this.symbolic` in the `if`.

PR #245, opened 2020-04-03
Bugfix: String-typed ConstantTerms for strings and constant symbols mistakenly equal (Issue #244)

madmike200590:
Fixes the bug described in issue #244
* Extended `ConstantTerm#equals` to check whether `this.symbolic != other.symbolic`
* Extended `ConstantTerm#compareTo` to compare based on priority if `this.symbolic != other.symbolic`
* additional Test Case in `TermTest` checking the condition from issue #244

### DIFF
--- a/src/test/java/at/ac/tuwien/kr/alpha/common/TermTest.java
+++ b/src/test/java/at/ac/tuwien/kr/alpha/common/TermTest.java
@@ -4,6 +4,8 @@ import at.ac.tuwien.kr.alpha.common.terms.ConstantTerm;
 import at.ac.tuwien.kr.alpha.common.terms.FunctionTerm;
 import at.ac.tuwien.kr.alpha.common.terms.Term;
 import at.ac.tuwien.kr.alpha.common.terms.VariableTerm;
+
+import org.junit.Assert;
 import org.junit.Test;
 
 import java.util.LinkedList;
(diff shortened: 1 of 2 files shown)


src/test/java/at/ac/tuwien/kr/alpha/common/TermTest.java
```diff
@@ -66,4 +69,19 @@ public void testTermOrdering() throws Exception {
...
+		ConstantTerm<String> constantSymbol = ConstantTerm.getSymbolicInstance("string");
+		// reference equality must hold for both constructs..
+		Assert.assertTrue(stringConstant == ConstantTerm.getInstance("string"));
+		ConstantTerm<String> sameConstantSymbol = ConstantTerm.getSymbolicInstance("string");
+		Assert.assertTrue(constantSymbol == sameConstantSymbol);
+		// ... but stringConstant and constantSymbol are NOT the same thing!
+		Assert.assertNotEquals(stringConstant.hashCode(), constantSymbol.hashCode());
+		Assert.assertNotEquals(stringConstant, constantSymbol);
+		// ... same goes for compareTo, must behave in sync with equals and hashCode
+		Assert.assertNotEquals(0, stringConstant.compareTo(constantSymbol));
```
AntoniusW, line 85: Two minor things: First, I am not sure if we are guaranteed that the compiler will replace both `"string"` objects with exactly the same object; this could be enforced by making it an extra variable `String str = "string";` and then constructing both contant terms using `str`. Secondly, please adapt the comment style to the usual one, i.e., full sentences with upper-case first word.

madmike200590: Both fixed. As far as I'm aware, string literals are compiled into being the same object, so as long as you don't create strings as `new String("stringLiteral")`, that should be fine - that being said, thanks for the heads-up, I changed the unit test to the bulletproof variant. Unfortunately I didn't find a good reference on Java's internal String handling in a quick google search.

PR #249, opened 2020-04-23
Throw exception if variable name is invalid

rtaupe:
Throw exception if variable name does not start with _ or upper-case letter (this is mainly to notice errors in unit tests!)

### DIFF
--- a/src/main/java/at/ac/tuwien/kr/alpha/common/terms/VariableTerm.java
+++ b/src/main/java/at/ac/tuwien/kr/alpha/common/terms/VariableTerm.java
@@ -1,3 +1,31 @@
+/*
+ * Copyright (c) 2016-2018, 2020, the Alpha Team.
+ * All rights reserved.
+ *
+ * Additional changes made by Siemens.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1) Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2) Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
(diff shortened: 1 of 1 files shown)


src/main/java/at/ac/tuwien/kr/alpha/common/terms/VariableTerm.java
```diff
@@ -23,6 +48,10 @@ private VariableTerm(String variableName) {
 	}
 
 	public static VariableTerm getInstance(String variableName) {
+		final char firstChar = variableName.charAt(0);
+		if (firstChar != ANONYMOUS_VARIABLE_PREFIX.charAt(0) && !Character.isUpperCase(firstChar)) {
+			throw new IllegalArgumentException("Variable name must start with upper-case letter or " + ANONYMOUS_VARIABLE_PREFIX);
+		}
```
madmike200590, line 54: I'm assuming this can only ever be triggered from unit tests (since anything triggering the `if` would be taken care of by the parser normally), so I'm not sure this check should be here. At least, I think there should be a comment explaining it's purpose - for someone "naively" reading the code it might look like a syntax check for parsed programs otherwise..

Alternatively, would your problem be solved by writing a utlity method for unit tests? something like the below snippet, that can be used in tests, but does not "bloat" the actual production code:
```
public static VariableTerm generateVariableTerm(String varName){
    if(!varName.maches("(_)|([A-Z][A-Za-z0-9_]*))){
        throw .....
    }
    return new VariableTerm(varName);
}

AntoniusW:
This PR might need some more explanation. In what way does a check like this help?
Currently, we sometimes create internal variables and have them named such that their name is invalid with respect to the ASP core standard, as this guarantees that they cannot come into conflict with any variables of the input program. This check now restricts variable names but still allows that variables are named `_foo`, which is against the standard. Why this restriction and not another?

rtaupe:
The motivation for this PR was that I accidentally used `VariableTerm.getInstance("x")` instead of `ConstantTerm.getInstance("x")`, which led to confusing results.
On second thought, I have realised that variable names starting with upper-case letters are that way purely because of convention, and there is no technical reason for such a restriction inside the solver. Therefore, I´m closing this PR.

Issue #234, opened 2020-03-09
Negation not taken into account for external atoms


madmike200590:
In the program below, the negation on the external atom `&stdlib_string_matches_regex[S, ".*foo.*"]` is ignored by Alpha.

```
string("bla").
string("blubb").
string("foo").
string("bar").
{ strcat(S1, S2) } :- string(S1), string(S2).
resultstring(SCAT) :- strcat(S1, S2), &stdlib_string_concat[S1, S2](SCAT).
:- resultstring(S), &stdlib_string_length[S](LEN), LEN != 6.
:- resultstring(S), not &stdlib_string_matches_regex[S, ".*foo.*"].
has_resultstring :- resultstring(_).
:- not has_resultstring.
```
Currently, answer sets (filtered for `resultstring/1`) are:
```
resultstring(blabar).
resultstring(blabla).
resultstring(barbar). 
resultstring(barbla). 
```
... i.e., strings the constraint in question kills strings *containing* "foo", rather than strings *not containing* "foo".

A test case illustrating the problem is committed in branch `negated_externals` (`AspStandardLibrary.negatedExternal`)

PR #252, opened 2020-05-06
Fix evaluation of negated external atoms (Issue #234)

madmike200590:
Fixes the issue described in #234.

For external atoms that have outputs, if the literal is negated, the truth value of the literal is determined as follows:
```The (negated external) literal lit is true iff no list of terms TL exists in the external atom's output, such that the terms in TL equal the substituted values of the literal's output variables (i.e. the values those variables are bound to by the grounder)```

Also, `FixedInterpretationLiteral#getSubstitutions` renamed to `getSatisfyingSubstitutions` to properly represent method semantics in the method name.

### DIFF
--- a/src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalLiteral.java
+++ b/src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalLiteral.java
@@ -157,7 +157,7 @@ public class ExternalLiteral extends FixedInterpretationLiteral {
 	 * 
 	 * @param externalMethodResult The term lists obtained from evaluating the external atom
 	 *                (i.e. calling the java method) encapsulated by this literal
-	 * @return true iff no list in output equals the external atom's output term
+	 * @return true iff no list in externalMethodResult equals the external atom's output term
 	 *         list as substituted by the grounder, false otherwise
 	 */
 	private boolean isNegatedLiteralSatisfied(Set<List<ConstantTerm<?>>> externalMethodResult) {
--- a/src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalLiteral.java
+++ b/src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalLiteral.java
@@ -148,7 +148,10 @@ public class ExternalLiteral extends FixedInterpretationLiteral {
 			}
 		} else {
 			if (this.isNegated()) {
-				return Collections.emptyList();
+				List<Substitution> retVal = new ArrayList<>();
+				retVal.add(partialSubstitution);
+				return retVal;
+				// return Collections.emptyList();
 			} else {
 				return this.buildSubstitutionsForOutputs(partialSubstitution, results);
 			}
--- a/src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalLiteral.java
+++ b/src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalLiteral.java
@@ -119,7 +119,7 @@ public class ExternalLiteral extends FixedInterpretationLiteral {
 	}
 
 	@Override
-	public List<Substitution> getSubstitutions(Substitution partialSubstitution) {
+	public List<Substitution> getSatisfyingSubstitutions(Substitution partialSubstitution) {
 		List<Term> input = getAtom().getInput();
 		List<Term> substitutes = new ArrayList<>(input.size());
 
@@ -156,6 +156,23 @@ public class ExternalLiteral extends FixedInterpretationLiteral {
 				return this.buildSubstitutionsForOutputs(partialSubstitution, results);
 			}
 		}
+	}
+
+	/**
+	 * Checks whether this (negated) external literal is satisfied under the given
+	 * partial substitution.
+	 * Note that this method is only called on negated external literals that have
+	 * outputs. In that specific case, the literal itself does not bind any
+	 * variables, i.e. the literal is satisfied iff the output terms obtained
+	 * by evaluating the underlying external atom match the values to which the
+	 * respective output variables are bound in the partial substitution.
+	 * 
+	 * @param partialSubstitution
+	 * @param outputs
+	 * @return
+	 */
+	private boolean checkSatisfied(Substitution partialSubstitution, Set<List<ConstantTerm<?>>> outputs) {
+		return false;
 	}
 
 	private List<Substitution> buildSubstitutionsForOutputs(Substitution partialSubstitution, Set<List<ConstantTerm<?>>> outputs) {
--- a/src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalLiteral.java
+++ b/src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalLiteral.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2017-2018, the Alpha Team.
+ * Copyright (c) 2017-2020, the Alpha Team.
  * All rights reserved.
  * 
  * Additional changes made by Siemens.
@@ -27,14 +27,16 @@
  */
 package at.ac.tuwien.kr.alpha.common.atoms;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
 import at.ac.tuwien.kr.alpha.common.terms.ConstantTerm;
 import at.ac.tuwien.kr.alpha.common.terms.Term;
 import at.ac.tuwien.kr.alpha.common.terms.VariableTerm;
 import at.ac.tuwien.kr.alpha.grounder.Substitution;
-
-import java.util.*;
-
-import static java.util.Collections.emptyList;
 
 /**
  * Contains a potentially negated {@link ExternalAtom}.
@@ -44,14 +46,15 @@ public class ExternalLiteral extends FixedInterpretationLiteral {
 	public ExternalLiteral(ExternalAtom atom, boolean positive) {
 		super(atom, positive);
 	}
-	
+
 	@Override
 	public ExternalAtom getAtom() {
-		return (ExternalAtom)atom;
+		return (ExternalAtom) atom;
 	}
 
 	/**
-	 * Returns a new copy of this literal whose {@link Literal#isNegated()} status is inverted
+	 * Returns a new copy of this literal whose {@link Literal#isNegated()} status
+	 * is inverted
 	 */
 	@Override
 	public ExternalLiteral negate() {
@@ -68,14 +71,15 @@ public class ExternalLiteral extends FixedInterpretationLiteral {
 
 	@Override
 	public Set<VariableTerm> getBindingVariables() {
-		// If the external atom is negative, then all variables of input and output are non-binding
+		// If the external atom is negative, then all variables of input and output are
+		// non-binding
 		// and there are no binding variables (like for ordinary atoms).
 		// If the external atom is positive, then variables of output are binding.
 
 		if (!positive) {
 			return Collections.emptySet();
 		}
-		
+
 		List<Term> output = getAtom().getOutput();
 
 		Set<VariableTerm> binding = new HashSet<>(output.size());
@@ -93,7 +97,7 @@ public class ExternalLiteral extends FixedInterpretationLiteral {
 	public Set<VariableTerm> getNonBindingVariables() {
 		List<Term> input = getAtom().getInput();
 		List<Term> output = getAtom().getOutput();
-		
+
 		// External atoms have their input always non-binding, since they cannot
 		// be queried without some concrete input.
 		Set<VariableTerm> nonbindingVariables = new HashSet<>();
@@ -101,7 +105,8 @@ public class ExternalLiteral extends FixedInterpretationLiteral {
 			nonbindingVariables.addAll(term.getOccurringVariables());
 		}
 
-		// If the external atom is negative, then all variables of input and output are non-binding.
+		// If the external atom is negative, then all variables of input and output are
+		// non-binding.
 		if (!positive) {
 			for (Term out : output) {
 				if (out instanceof VariableTerm) {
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalLiteral.java [hunk 1]
```diff
@@ -68,14 +71,15 @@ public ExternalLiteral substitute(Substitution substitution) {
 
 	@Override
 	public Set<VariableTerm> getBindingVariables() {
-		// If the external atom is negative, then all variables of input and output are non-binding
+		// If the external atom is negative, then all variables of input and output are
+		// non-binding
 		// and there are no binding variables (like for ordinary atoms).
 		// If the external atom is positive, then variables of output are binding.
```
AntoniusW, line 77: Please either reformat all, or none of those lines.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalLiteral.java
```diff
@@ -114,35 +119,82 @@ public ExternalLiteral substitute(Substitution substitution) {
...
 
 	@Override
-	public List<Substitution> getSubstitutions(Substitution partialSubstitution) {
+	public List<Substitution> getSatisfyingSubstitutions(Substitution partialSubstitution) {
 		List<Term> input = getAtom().getInput();
-		List<Term> output = getAtom().getOutput();
-		List<Substitution> substitutions = new ArrayList<>();
 		List<Term> substitutes = new ArrayList<>(input.size());
 
+		// evaluate the external atom
```
AntoniusW, line 126: Just nitpicking here: please have a full sentence with proper capitalisation here.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalLiteral.java
```diff
@@ -114,35 +119,82 @@ public ExternalLiteral substitute(Substitution substitution) {
...
+	 * the respective output variables are bound. The result of this method assumes
+	 * that the literal is negated, i.e. for the underlying atom AT, it represents
+	 * the truth value !AT.
+	 * Furthermore, this method assumes that the output terms of the underlying atom
+	 * are {@link ConstantTerm}s, i.e. that the grounder's current partial
+	 * substitution has already been applied to them. That assumption is safe since
+	 * in case of a negated external literal, the output variables are always
+	 * non-binding.
+	 * 
+	 * @param outputs The term lists obtained from evaluating the external atom
```
AntoniusW, line 158: I suggest to rename this parameter, maybe to something like `externalAtomResults` or something else indicating this is not the output of the method.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalLiteral.java
```diff
@@ -114,35 +119,82 @@ public ExternalLiteral substitute(Substitution substitution) {
...
+	 *                (i.e. calling the java method) encapsulated by this literal
+	 * @return true iff no list in output equals the external atom's output term
+	 *         list as substituted by the grounder, false otherwise
+	 */
+	private boolean checkNegatedLiteralSatisfied(Set<List<ConstantTerm<?>>> outputs) {
```
AntoniusW, line 163: I think `isNegatedLiteralSatisfied` would be a better name for this method.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalLiteral.java
```diff
@@ -114,35 +119,82 @@ public ExternalLiteral substitute(Substitution substitution) {
...
+		for (List<ConstantTerm<?>> output : outputs) {
+			outputMatches = true;
+			for (int i = 0; i < externalAtomOutTerms.size(); i++) {
+				if (!output.get(i).equals(externalAtomOutTerms.get(i))) {
+					outputMatches = false;
+					break;
+				}
 			}
+			if (outputMatches) {
+				// we found one term list where all terms match the ground output terms of the
```
AntoniusW, line 175: Capitalisation.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalLiteral.java
```diff
@@ -114,35 +119,82 @@ public ExternalLiteral substitute(Substitution substitution) {
...
+				// external atom, therefore the atom is true and the (negative) literal false.
+				return false;
+			}
+		}
+		// we checked all term list and none matches the ground output terms, therefore
```
AntoniusW, line 180: Capitalisation.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalLiteral.java
```diff
@@ -114,35 +119,82 @@ public ExternalLiteral substitute(Substitution substitution) {
...
+		// the external atom is false, making the literal true
```
AntoniusW, line 181: `.`

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalLiteral.java
madmike200590: fixed

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalLiteral.java
madmike200590: fixed

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalLiteral.java
madmike200590: It actually is the output of the method - but renaming definitely makes sense. It's called `externalMethodResult` now.

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalLiteral.java
madmike200590: done

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalLiteral.java
madmike200590: fixed

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalLiteral.java
madmike200590: fixed

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalLiteral.java
madmike200590: fixed

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalLiteral.java
```diff
@@ -114,35 +119,82 @@ public ExternalLiteral substitute(Substitution substitution) {
...
+					outputMatches = false;
+					break;
+				}
 			}
+			if (outputMatches) {
+				// we found one term list where all terms match the ground output terms of the
+				// external atom, therefore the atom is true and the (negative) literal false.
+				return false;
+			}
+		}
```
AntoniusW: Using labelled continue, the double-loop would look as follows:

```suggestion
		atomTuplesLoop:
		for (List<ConstantTerm<?>> output : outputs) {
			for (int i = 0; i < externalAtomOutTerms.size(); i++) {
				if (!output.get(i).equals(externalAtomOutTerms.get(i))) {
					continue atomTuplesLoop;
				}
			}
			// Found one term list where all terms match the ground output terms of the
			// external atom, therefore the atom is true and the (negative) literal false.
			return false;
		}
```

madmike200590: Personally, I don't think the labelled continue improves readability. I know this is a matter of personal taste, but to me it just looks "weird"... So if you're ok with it, I'd rather refrain from using that...

src/test/java/at/ac/tuwien/kr/alpha/api/externals/stdlib/AspStandardLibraryTest.java
```diff
@@ -139,4 +156,21 @@ public void programWithStringStuff() throws IOException {
 		}
 	}
 
+	@Test
+	@SuppressWarnings("unchecked")
+	public void negatedExternal() throws IOException {
+		Alpha alpha = new Alpha();
+		Program prog = alpha.readProgram(InputConfig.forString(NEGATED_EXTERNAL_ASP));
+		Set<AnswerSet> answerSets = alpha.solve(prog).collect(Collectors.toSet());
+		// Verify every result string has length 6 and contains "foo"
```
AntoniusW, line 165: Also verify that there is the expected number of answer sets!

src/test/java/at/ac/tuwien/kr/alpha/common/fixedinterpretations/FixedInterpretationLiteralsTest.java
```diff
@@ -0,0 +1,199 @@
...
+	@at.ac.tuwien.kr.alpha.api.externals.Predicate
+	public static boolean isOne(int value) {
+		return value == 1;
+	}
+
+	/**
+	 * Dummy method to test external atoms with more than one output term list.
+	 * Dummy input param which is not used exists solely to avoid producing an edge
+	 * case
+	 * where there is output, but no input
```
AntoniusW, line 32: Formatting.

src/test/java/at/ac/tuwien/kr/alpha/api/externals/stdlib/AspStandardLibraryTest.java
```diff
@@ -139,4 +156,21 @@ public void programWithStringStuff() throws IOException {
...
+		for (AnswerSet as : answerSets) {
+			for (Atom atom : as.getPredicateInstances(Predicate.getInstance("resultstring", 1))) {
+				String resultstring = ((ConstantTerm<String>) atom.getTerms().get(0)).getObject();
+				LOGGER.info("ResultSttring is {}", resultstring);
```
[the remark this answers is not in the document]
madmike200590: no good reason for that, changed to debug

src/test/java/at/ac/tuwien/kr/alpha/common/fixedinterpretations/FixedInterpretationLiteralsTest.java
```diff
@@ -0,0 +1,199 @@
...
+			+ "positive_external_with_output_dontfire :- "
+			+ "&stdlib_datetime_parse[\"01.01.2121 21:11:12\", \"dd.MM.yyyy HH:mm:ss\"](Y, MO, D, H, MI, S), Y = 2345, MO = 1, D = 1, H = 21, MI = 11, S = 12."
+			+ "negative_external_with_output :- Y = 2121, MO = 1, D = 1, H = 21, MI = 11, S = 12, "
+			+ "not &stdlib_datetime_parse[\"22.02.2222 21:11:12\", \"dd.MM.yyyy HH:mm:ss\"](Y, MO, D, H, MI, S)."
+			+ "negative_external_with_output_dontfire :- Y = 2222, MO = 2, D = 22, H = 21, MI = 11, S = 12,"
+			+ "not &stdlib_datetime_parse[\"22.02.2222 21:11:12\", \"dd.MM.yyyy HH:mm:ss\"](Y, MO, D, H, MI, S)."
+			+ "negative_external_multioutput :- START = \"Graz\", END = \"Salzburg\", not &connection[train_network](START, END)."
+			+ "negative_external_multioutput_dontfire :- START = \"Klagenfurt\", END = \"Villach\", not &connection[train_network](START, END)."
+			+ "positive_external_multioutput :- START = \"Klagenfurt\", END = \"Villach\", &connection[train_network](START, END)."
+			+ "positive_external_multioutput_dontfire :- START = \"Graz\", END = \"Salzburg\", &connection[train_network](START, END).";
```
AntoniusW, line 67: This is very good and tests many combinations!
I think, an important case is missing, however: one where the external literal is guaranteed to generate variable bindings (here, the evaluation order could always bind the output variables to some constant, before evaluating the external literal. So, I think a rule where some variable outputs from the external literal are fed into the head of the rule should be added (and checked). For example:
`"external_literal_output(START, END) :-  &connection[train_network](START, END)."`

src/test/java/at/ac/tuwien/kr/alpha/common/fixedinterpretations/FixedInterpretationLiteralsTest.java
```diff
@@ -0,0 +1,199 @@
...
+		this.externals.putAll(Externals.scan(AspStandardLibrary.class));
+		this.externals.putAll(Externals.scan(FixedInterpretationLiteralsTest.class));
+	}
+
+	@Test
+	public void positiveNumericComparison() {
+		Optional<AnswerSet> answer = this.alpha.solve(this.alpha.readProgramString(TEST_PROG, this.externals)).findFirst();
+		if (!answer.isPresent()) {
+			throw new IllegalStateException("Test setup failed, no answer set!");
+		}
```
AntoniusW, line 84: Just curious: why is this not an `assertTrue(answer.isPresent());" ?

src/test/java/at/ac/tuwien/kr/alpha/api/externals/stdlib/AspStandardLibraryTest.java
madmike200590: fixed

src/test/java/at/ac/tuwien/kr/alpha/common/fixedinterpretations/FixedInterpretationLiteralsTest.java
madmike200590: fixed

src/main/java/at/ac/tuwien/kr/alpha/common/atoms/ExternalLiteral.java
AntoniusW: I am fine with that. As we discussed, I think the labelled continue is actually easier to read and so I wanted to at least show it. :-)

src/test/java/at/ac/tuwien/kr/alpha/common/fixedinterpretations/FixedInterpretationLiteralsTest.java
madmike200590: Because in the very first version of this class, the whole solving part was in the constructor of the test class (since it's identical for every test case). I only changed that later on when I saw that I couldn't easily step through that code with a debugger when it was part of the constructor. As it stands right now, there's no good reason no to do the assert variant, so I changed that now.

src/test/java/at/ac/tuwien/kr/alpha/common/fixedinterpretations/FixedInterpretationLiteralsTest.java
madmike200590: Thanks for catching that! Added a test case to cover that..

PR #257, opened 2020-07-01
Update most recent Java version and Gradle version

lorenzleutgeb:
Java 13 [went out of "Premier Support" in March](https://www.oracle.com/java/technologies/java-se-support-roadmap.html), therefore I removed it. Also, there is now Java 14, which is not yet succeeded by a more recent release, so I added it.


lorenzleutgeb:
Changes `gradle/*` and `gradlew*` were automatically generated by `./gradlew wrapper`.

PR #262, opened 2020-07-22
Command Line Interface - Make separator string for atoms within printed answer set configurable

madmike200590:
Adds a new command line parameter (`sep <separator>`) in order to influence the way answer sets are printed to `stdout`.

Example:
```
$ java -jar Alpha-bundled.jar -str "a. b(c). d(e, f)." -sep ",\n"
```
formats the answer set as:
```
{ a,
b(c),
d(e, f) }
```

### DIFF
--- a/src/main/java/at/ac/tuwien/kr/alpha/common/BasicAnswerSetFormatter.java
+++ b/src/main/java/at/ac/tuwien/kr/alpha/common/BasicAnswerSetFormatter.java
@@ -1,0 +1,33 @@
+package at.ac.tuwien.kr.alpha.common;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.SortedSet;
+import java.util.stream.Collectors;
+
+import at.ac.tuwien.kr.alpha.common.atoms.Atom;
+
+public class BasicAnswerSetFormatter implements AnswerSetFormatter<String> {
+
+	private final String atomSeparator;
+
+	public BasicAnswerSetFormatter(String atomSeparator) {
+		this.atomSeparator = atomSeparator;
+	}
+
+	@Override
+	public String format(AnswerSet answerSet) {
+		List<String> predicateInstanceStrings = new ArrayList<>();
+		for (Predicate p : answerSet.getPredicates()) {
+			SortedSet<Atom> instances;
+			if ((instances = answerSet.getPredicateInstances(p)) == null || instances.isEmpty()) {
+				predicateInstanceStrings.add(p.getName());
+			} else {
+				List<String> atomStrings = instances.stream().map((atom) -> atom.toString()).collect(Collectors.toList());
+				predicateInstanceStrings.add(String.join(this.atomSeparator, atomStrings));
+			}
+		}
+		return "{ " + String.join(this.atomSeparator, predicateInstanceStrings) + " }";
+	}
+
+}
--- a/src/main/java/at/ac/tuwien/kr/alpha/config/CommandLineParser.java
+++ b/src/main/java/at/ac/tuwien/kr/alpha/config/CommandLineParser.java
@@ -27,8 +27,13 @@
  */
 package at.ac.tuwien.kr.alpha.config;
 
-import at.ac.tuwien.kr.alpha.solver.BinaryNoGoodPropagationEstimation;
-import at.ac.tuwien.kr.alpha.solver.heuristics.BranchingHeuristicFactory.Heuristic;
+import java.io.ByteArrayOutputStream;
+import java.io.FileInputStream;
+import java.io.PrintWriter;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Consumer;
+
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.DefaultParser;
 import org.apache.commons.cli.HelpFormatter;
@@ -36,15 +41,12 @@ import org.apache.commons.cli.Option;
 import org.apache.commons.cli.Options;
 import org.apache.commons.cli.ParseException;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.text.StringEscapeUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.ByteArrayOutputStream;
-import java.io.FileInputStream;
-import java.io.PrintWriter;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.function.Consumer;
+import at.ac.tuwien.kr.alpha.solver.BinaryNoGoodPropagationEstimation;
+import at.ac.tuwien.kr.alpha.solver.heuristics.BranchingHeuristicFactory.Heuristic;
 
 /**
  * Parses given argument lists (as passed when Alpha is called from command line) into {@link AlphaConfig}s and {@link InputConfig}s.
@@ -417,7 +419,7 @@ public class CommandLineParser {
 	}
 
 	private void handleAtomSeparator(Option opt, SystemConfig cfg) {
-		cfg.setAtomSeparator(opt.getValue(SystemConfig.DEFAULT_ATOM_SEPARATOR));
+		cfg.setAtomSeparator(StringEscapeUtils.unescapeJava(opt.getValue(SystemConfig.DEFAULT_ATOM_SEPARATOR)));
 	}
 	
 }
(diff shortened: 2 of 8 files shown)


AntoniusW:
I think this is a useful addition, but it is not fully clear to me why only the atom separator is configurable?
And I am also not sure if more configurability here is over-engineering. What do you think? Can you explain the rationale behind having exactly the proposed functionality?

One might also argue that there are better programs to format strings (e.g. sed/grep), what is the argument for having that functionality directly in the solver?

(NB: the tests are failing because of checkstyle complaining.)

madmike200590:
> One might also argue that there are better programs to format strings (e.g. sed/grep), what is the argument for having that functionality directly in the solver?

sed and grep are both great, however, my motivation for this was that I simply needed every atom on a new line (in order to grep for something), and hit a dead end trying to achieve that with sed. The reason for that is that replacing the string `), ` with `),\n` doesn't work when you have atoms with function terms in the answer set, such as `p(q, f(x, y), z), p(r, (f(u, v), z), ...`, so I think having this in the solver is a nice convenience feature with the potential to make working with the solver a lot more comfortable.
This might be even more useful to people working on Windows, where string manipulation in the shell is more awkward.

> (NB: the tests are failing because of checkstyle complaining.)

Checkstyle warnings are fixed now - sorry for that, it was a bit of a rush job...

> I think this is a useful addition, but it is not fully clear to me why only the atom separator is configurable?
> And I am also not sure if more configurability here is over-engineering. What do you think? Can you explain the rationale behind having exactly the proposed functionality?

Actually, initially, I wanted to do something like a full format string, allowing things like
```
java -jar Alpha-bundled.jar -i someFile.asp -fmt "Instance of %(predicate) of arity %(arity): (%(terms))\n"
```
i.e. basically doing something "printf-like" with named parameters internally. However, I think spending that much time on something relatively minor didn't seem justified. Also, I think this would be relatively complicated for users, so I decided against such a solution.

src/main/java/at/ac/tuwien/kr/alpha/config/CommandLineParser.java
```diff
@@ -125,7 +127,11 @@
 			.build();
 	private static final Option OPT_GROUNDER_ACCUMULATOR_ENABLED = Option.builder("acc").longOpt("enableAccumulator")
 			.desc("activates the accumulator grounding strategy by disabling removal of instances from grounder memory in certain cases (default: "
-						+ SystemConfig.DEFAULT_GROUNDER_ACCUMULATOR_ENABLED + ")")
+					+ SystemConfig.DEFAULT_GROUNDER_ACCUMULATOR_ENABLED + ")")
+			.build();
+	private static final Option OPT_OUTPUT_ATOM_SEPARATOR = Option.builder("sep").longOpt("atomSeparator").hasArg(true).argName("separator")
+			.desc("a character (sequence) to use s separator for atoms in printed answer sets. (default: "
```
AntoniusW, line 133:
```suggestion
			.desc("a character (sequence) to use as separator for atoms in printed answer sets (default: "
```

src/test/java/at/ac/tuwien/kr/alpha/config/CommandLineParserTest.java
```diff
@@ -51,22 +51,22 @@
 	public void help() throws ParseException {
 		StringBuilder bld = new StringBuilder();
 		CommandLineParser parser = new CommandLineParser(DEFAULT_COMMAND_LINE, (msg) -> bld.append(msg));
-		parser.parseCommandLine(new String[] {"-h"});
+		parser.parseCommandLine(new String[] {"-h" });
```
AntoniusW, line 54: This is strange behaviour of your auto-formatter: why put the whitespace behind the string but not in front of it? Please disable this and revert to the previous state on all locations throughout this file, which is more consistent.

madmike200590: Actually this is a weird quirk of our checkstyle config: My formatter formats as `parser.parseCommandLine(new String[] { "-h" });` i.e. with whitespace both at start and end, however, checkstyle complains about the whitespace after `{`.
Personally, I disagree with checkstyle and would be in favor of following my formatter ;-)

AntoniusW: Well, since the project has been going with no whitespace there, I'd ask to keep in line with that.

lorenzleutgeb:
I also think that some caution is warranted here. The CLI should be slim, and this option seems a little too low-level (see also [the Unix philosophy](https://en.wikipedia.org/wiki/Unix_philosophy)). What is the motivation for the proposed change?

If it is to allow easier processing of results, then it might be worth considering an even simpler line/table-based format (nice to `grep`), or even JSON (with shell tools like [`jq`](https://stedolan.github.io/jq/) for scripting).

src/main/java/at/ac/tuwien/kr/alpha/common/BasicAnswerSetFormatter.java
```diff
@@ -0,0 +1,33 @@
...
+
+	private final String atomSeparator;
+
+	public BasicAnswerSetFormatter(String atomSeparator) {
+		this.atomSeparator = atomSeparator;
+	}
+
+	@Override
+	public String format(AnswerSet answerSet) {
+		List<String> predicateInstanceStrings = new ArrayList<>();
```
AntoniusW, line 20: There is a lot of `String`-creation and copying going on here for printing an answer set. I thought for a while on how to avoid that, but it already starts at `Atom.toString()`, so in order to get rid of the string-copying, one would need to start there and add facilities to write an `Atom` to a `StringBuilder`.
(Just mentioning this as potential future optimisation.)

AntoniusW:
@lorenzleutgeb what do you mean by "an even simpler line/table-based format"?
Writing JSON seems a bit too much for the moment, though given the right interfaces in the solver, it would be nice to have.

lorenzleutgeb:
> @lorenzleutgeb what do you mean by "an even simpler line/table-based format"?
> Writing JSON seems a bit too much for the moment, though given the right interfaces in the solver, it would be nice to have.

I didn't have any particular format in mind. If one wants to `grep` through answer sets, maybe just split atoms by `\n`, terms by `\t` and answer sets by an additional empty line. Suppose that this would be enabled as `--print-table` or similar.

```
a
b       c
b       e
d       e       f

a
b       e
b       c
d       c       f
```

One might do:
 * `java -jar alpha.jar ... | grep d | cut -f2`
 * `java -jar alpha.jar ... | grep --context=10 c`

This obviously deviates a lot from the common notation with `{`/`}` and `p(x,y)` (parentheses and comma).

However, adopting a common structured notation like JSON could be a quick win. There are many parsers and tools available for "post processing" the output and integrating Alpha with other systems.

Note that these two approaches are by no means mutually exclusive.

src/test/java/at/ac/tuwien/kr/alpha/config/CommandLineParserTest.java
madmike200590: fixed

madmike200590:
> I also think that some caution is warranted here. The CLI should be slim, and this option seems too low-level
Yes, definitely. After discussing this with @AntoniusW, we came to the conclusion that, ideally we'd be able to choose an `AnswerSetFormatter` via commandline-argument and the supply it with it's formatter-specific options, thereby gaining flexibility and transparency.
However, trying to quickly implement that on top of the current - already a bit overloaded - CLI would make matters even worse in my opinion. I created a new issue (#264) summarizing the problem and as the place to do a proper overhaul of the CLI.

> However, adopting a common structured notation like JSON could be a quick win. There are many parsers and tools available for "post processing" the output and integrating Alpha with other systems.
Once again, I totally agree, but I think this is dependent on issue #264. I created #265 specifically to start a discussion on which output formatting capabilities we want to have going forward.
While things like `printf`-style formatting and JSON definitely make sense, I'd also like to point out that, when using Alpha as a component of a larger application, the java API would be the way to go - communicating with the solver through `stdout` and `stdin` is probably mostly relevant when running benchmarks or quickly testing something.

> This PR seems to be rushed, as @madmike200590 admitted, and getting it in will lead to further breaking changes down the road when more complex output configuration is required.
I cannot disagree ;-) but would still like to get this PR merged as-is, simply because it's a small thing, and doing something "proper" here would be significantly more work - which I'm definitely in favor of investing, but I'd rather not block this PR because of it.

madmike200590:
@AntoniusW unless there are still unaddressed issues, could you please merge this PR?

AntoniusW (approved):

PR #270, opened 2020-10-01
Substitution lazy cloning

AntoniusW:
Removes the need for upfront cloning of Substitutions when checking unification.

### DIFF
--- a/src/main/java/at/ac/tuwien/kr/alpha/grounder/Substitution.java
+++ b/src/main/java/at/ac/tuwien/kr/alpha/grounder/Substitution.java
@@ -35,6 +35,7 @@ import at.ac.tuwien.kr.alpha.common.terms.Term;
 import at.ac.tuwien.kr.alpha.common.terms.VariableTerm;
 import at.ac.tuwien.kr.alpha.grounder.parser.ProgramPartParser;
 
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
@@ -45,6 +46,12 @@ import static at.ac.tuwien.kr.alpha.Util.oops;
 public class Substitution {
 
 	private static final ProgramPartParser PROGRAM_PART_PARSER = new ProgramPartParser();
+	public static final Substitution EMPTY_SUBSTITUTION = new Substitution() {
+		@Override
+		public <T extends Comparable<T>> Term put(VariableTerm variableTerm, Term groundTerm) {
+			throw oops("Should not be called on EMPTY_SUBSTITUTION");
+		}
+	};
 
 	protected TreeMap<VariableTerm, Term> substitution;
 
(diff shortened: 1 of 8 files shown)


src/main/java/at/ac/tuwien/kr/alpha/grounder/Substitution.java
```diff
@@ -68,29 +75,98 @@ public static Substitution unify(Literal literal, Instance instance, Substitutio
...
+			if (returnSubstitution == null) {
+				// All terms unify but there was no need to assign a new variable, return the input substitution.
+				return partialSubstitution;
 			}
-			return null;
+			return returnSubstitution;
 		}
-		return substitution;
+
+		boolean unifyTerms(Term termNonGround, Term termGround, Substitution partialSubstitution) {
```
madmike200590, line 98: It seems wrong (or at least really confusing) to me that we now have `UnificationHelper::unifyTerms(Term, Term, Substitution)` as well as `Substitution::unifyTerms(Term, Term)`. The only difference seems to be that the one in `Substitution` always operates on `this` and modifying the substitution it's called on, while the other one operates on the given `partialSubstitution`. I think the version in `UnificationHelper` should be adapted so that the other one isn't needed.

src/main/java/at/ac/tuwien/kr/alpha/grounder/Substitution.java
```diff
@@ -68,29 +75,98 @@ public static Substitution unify(Literal literal, Instance instance, Substitutio
...
+	 * If unification succeeds the unifying substitution is returned. The partial substitution is left unchanged.
+	 *
+	 * @param atom         the (potentially nonground) atom to unify.
+	 * @param instance     the ground instance to unify with.
+	 * @param substitution the (partial) substitution for the atom. This is left unchanged in all cases.
+	 * @return null if the unification fails, otherwise it is a unifying substitution. If the parameter substitution
+	 * 	already is a unifier, it is returned. If the unifying substitution is an extension of the input
+	 * 	substitution, a new substitution will be returned.
+	 */
+	public static Substitution unify(Atom atom, Instance instance, Substitution substitution) {
```
madmike200590, line 157: More of a general question, and maybe out of the scope of this review... We also have classes `Unifier` and `Unification`, with `Uninfication` implementing utility functions looking very similar to some of the methods in `Substitution`. In order to avoid duplicating code, or writing several very similar versions, would it make sense to use the (accordingly adapted) methods from `Unification` here?

src/main/java/at/ac/tuwien/kr/alpha/grounder/Substitution.java
```diff
@@ -68,29 +75,98 @@ public static Substitution unify(Literal literal, Instance instance, Substitutio
...
+	 * Helper class to lazily clone the input substitution of Substitution.unify only when needed.
 	 */
-	public static Substitution unify(Atom atom, Instance instance, Substitution substitution) {
-		for (int i = 0; i < instance.terms.size(); i++) {
-			if (instance.terms.get(i) == atom.getTerms().get(i) || substitution.unifyTerms(atom.getTerms().get(i), instance.terms.get(i))) {
-				continue;
+	private static class UnificationHelper {
+		Substitution returnSubstitution;
+
+		Substitution unify(List<Term> termList, Instance instance, Substitution partialSubstitution) {
```
madmike200590, line 83: Note that this is a matter of personal taste, but I'm not sure I agree with the return types here..
I think it might make sense if `unifyTerms` returned something other than boolean, thereby getting rid of the side-effects of potentially modifying the given partialSubstitution. Instead, `unifyTerms` could return a `Substitution` that may be either a new (cloned) one, or the one passed in in case nothing changed. Alternatively, we could also use an `Optional<Substitution>` here, where the presence of a value would indicate that a new variable got bound. The version with `Optional` could also be used for `Substitution::unify` as an alternative to returning `null`.

madmike200590 (changes requested):
I think everything is ok from a functional point of view, i.e. the code does what it's supposed to do, however, I find it hard to read and confusing in places, see more detailed comments. I'm aware not all of those are easy to address (or even necessarily need to be addressed), but I think some discussion makes sense.

AntoniusW:
> I think everything is ok from a functional point of view, i.e. the code does what it's supposed to do, however, I find it hard to read and confusing in places, see more detailed comments. I'm aware not all of those are easy to address (or even necessarily need to be addressed), but I think some discussion makes sense.

That is exactly the reason why I like our four-eyes policy on pull requests. It forces one to deliver better code. Thank you for the comments!

src/main/java/at/ac/tuwien/kr/alpha/grounder/Substitution.java
AntoniusW: Agreed, the `Substitution::unifyTerms(Term, Term)` is only ever needed in unit tests, while the new one is not tested. Will adapt the tests to run on the real thing and get rid of the other.

src/main/java/at/ac/tuwien/kr/alpha/grounder/Substitution.java
AntoniusW: There is a slight difference between the two (and maybe we should adapt our naming): the one in `Substitution` is assuming that variables only ever occur on the left-hand side (i.e., the second parameter is always ground). The version in `Unification` really computes the unifier between two non-ground `Atom`s. Maybe we should rename the methods in `Substitution` to `instantiate()` instead of `unify()`.

src/main/java/at/ac/tuwien/kr/alpha/grounder/Substitution.java
AntoniusW: Letting `unifyTerms` return a `Substitution` is a good idea. I would not want to return an `Optional` however, since substitutions are used at the core of grounding and `Optional` is likely slower and creates more objects.

src/main/java/at/ac/tuwien/kr/alpha/grounder/Substitution.java
AntoniusW: I decided to rename `Substitution::unify` to `Substitution::specialize` as it specializes the given partial substitution to also unify the given atom and the instance.

src/main/java/at/ac/tuwien/kr/alpha/grounder/Substitution.java
AntoniusW: Actually, I decided to remove `unifyTerms` as it was only ever used in unit tests. These unit tests have been adapted to now test the new/renamed `specializeSubstitution`.

AntoniusW:
Comments have been addressed, please check if this is better now.

madmike200590 (approved):
Thanks for all the improvements, this looks really good now!

PR #273, opened 2020-10-26
Enable weak constraints

### LINKED ISSUE #272: Flag DebugEnableInternalChecks partially ignored
It seems that in the current master, the flag to enable the internal checks is only passed to the grounding components, but not to the solving ones. Although the `DefaultSolver` receives it via the `SystemConfig` parameter in the constructor, it does not pass it on to parts like the `ChoiceManager`.

AntoniusW:
Adds weak constraints to Alpha. The parser is able to read them and in their presence the solver switches to a branch-and-bound algorithm. Returned answer sets are adorned with a valuation (weights at different levels as specified by the weak constraints).


AntoniusW:
This PR should be ready to be reviewed now.

AntoniusW:
Now also with test passing. :-)

madmike200590:
I'm not yet through with the complete review, but I think I found a bug in one my tests:
The following program only has 4 answer sets with weak constraints:
```
{a; b; c}.

:~ a. [1]
:~ b. [2]
:~ c. [3]
```
Alpha output is:
```
michael@michael-laptop:~/asp_snippets/weak_constraints$ java -jar Alpha-bundled.jar -i wc-test.asp
  376 I lections.Reflections Reflections took 98 ms to scan 1 urls, producing 1 keys and 8 values  
Answer set 1:
{ a, c }
Optimization: [4@0]
Answer set 2:
{ c }
Optimization: [3@0]
Answer set 3:
{ a }
Optimization: [1@0]
Answer set 4:
{  }
Optimization: []
SATISFIABLE
```

Since I didn't specify any limit on answer sets computed, I'd expect to get all of them with their associated weigths. Is that a misunderstanding on my part?

The program without weak constraints, i.e. `{a; b; c}.` yields the expected 8 answer sets:
```
michael@michael-laptop:~/asp_snippets/weak_constraints$ java -jar Alpha-bundled.jar -i wc-test.asp
  399 I lections.Reflections Reflections took 110 ms to scan 1 urls, producing 1 keys and 8 values  
Answer set 1:
{ a, c }
Answer set 2:
{ c }
Answer set 3:
{ a }
Answer set 4:
{  }
Answer set 5:
{ a, b, c }
Answer set 6:
{ a, b }
Answer set 7:
{ b, c }
Answer set 8:
{ b }
SATISFIABLE
```
I haven't done any detailed analysis on this yet, but wanted to raise it anyway so you can take a look as well.

AntoniusW:
To me that looks perfectly fine. The solver sees the weak constraints and switches over to optimisation using a branch-and-bound algorithm. That means branches whose valuation is higher than the currently-best-known valuation are cut off immediately and not expanded. So one usually expects a sequence of answer-sets with increasingly better valuation (i.e., less violations). Since the last answer-set has zero violated weak constraints (hence a valuation of 0 at all levels), it is the best answer-set and the solver stops there.

Note that this behaviour is similar to what other ASP solvers do. Enumerating all answer-sets is usually infeasible when optimization is enabled and by using optimization (i.e., weak constraints) one tells the solver that one is interested in good answers only.

madmike200590:
> To me that looks perfectly fine. The solver sees the weak constraints and switches over to optimisation using a branch-and-bound algorithm. That means branches whose valuation is higher than the currently-best-known valuation are cut off immediately and not expanded. So one usually expects a sequence of answer-sets with increasingly better valuation (i.e., less violations). Since the last answer-set has zero violated weak constraints (hence a valuation of 0 at all levels), it is the best answer-set and the solver stops there.
> 
> Note that this behaviour is similar to what other ASP solvers do. Enumerating all answer-sets is usually infeasible when optimization is enabled and by using optimization (i.e., weak constraints) one tells the solver that one is interested in good answers only.

As per our discussion last week, I think we need the following changes:
- default behavior when Alpha is called without a specific number of answer sets to compute: Same as for programs without weak constraints, this should enumerate all answer sets along with their respective weights (needed for development/debugging)
- manual initial bound for answer set search: Add a commandline flag "-mw"/"--maxWeight" that, when used in conjuction with "-n" (number of answer sets) instructs Alpha to find n answer sets where no answer set must exceed the given weight (i.e. use the value of mw as initial bound for branch&bound)

madmike200590:
> As per our discussion last week, I think we need the following changes:
> 
> * default behavior when Alpha is called without a specific number of answer sets to compute: Same as for programs without weak constraints, this should enumerate all answer sets along with their respective weights (needed for development/debugging)
> * manual initial bound for answer set search: Add a commandline flag "-mw"/"--maxWeight" that, when used in conjuction with "-n" (number of answer sets) instructs Alpha to find n answer sets where no answer set must exceed the given weight (i.e. use the value of mw as initial bound for branch&bound)
> * bugfix: When Alpha finds an optimum solution, this should be reported textually (currently only says "SATISFIABLE", see above example)

@AntoniusW I just noticed that this is still tagged as pending my review - have the aforementioned changes been made, i.e. can we move this forward?

AntoniusW:
> 
> @AntoniusW I just noticed that this is still tagged as pending my review - have the aforementioned changes been made, i.e. can we move this forward?

@madmike200590, I think it is ready now. The functionality to not run optimization but still get valuations of answer sets should work now. An (exclusive) upper bound can also be set manually. Furthermore, current master has been merged and this branch now follows the modularization introduced in master. The code is ready for another round of reviews (and subsequent merging).

alpha-commons/src/main/java/at/ac/tuwien/kr/alpha/commons/WeightedAnswerSet.java
```diff
@@ -0,0 +1,154 @@
...
+		return joiner.toString();
+	}
+
+	/**
+	 * Extract weights at levels from string of comma-separated pairs of the form weight@level.
+	 * Multiple weights for the same level are summed-up.
+	 * @param weightAtLevels
+	 * @return a TreeMap containing
+	 */
+	public static TreeMap<Integer, Integer> weightPerLevelFromString(String weightAtLevels) {
```
madmike200590, line 131: This is the wrong place to parse command-line arguments. The parsing of this string should happen in `CommandLineParser`. Internally, weights per level should only be passed around as some kind of `Map`s.

madmike200590 (changes requested):
Looks good from my initial tests!
One small thing to fix - commandline argument parsing for weight/level pairs is a bit "over the place"

alpha-commons/src/main/java/at/ac/tuwien/kr/alpha/commons/WeightedAnswerSet.java
AntoniusW: This is a good catch and I was considering different options for where to put this code. The main problem is, that the same functionality is needed in `WeakConstraintsTests` in the tests of the core module. Relying on code from the cli-app module there felt like a bad idea and the other option would be having the code duplicated. What would you suggest for this kind of dependency?

PR #274, opened 2020-10-27
Modules

lorenzleutgeb:
# Codebase Modularization and API Streamlining

The idea of this PR is to make development of Alpha as well as applications using Alpha less cumbersome by partitioning the existing code into smaller, easier to understand packages, as well as providing a better public API for applications using Alpha.

## Modules

The codebase is divided into the following modules: alpha-api, alpha-commons, alpha-core, alpha-solver and alpha-cli-app.

### Module alpha-api

The api module consists mainly of interfaces specifying the objects API consumers (i.e. applications using Alpha) interact with, such as programs, rules, atoms, literals and answer sets. It also holds implementations of some configuration classes that are used across modules.

The rationale for specifying all API objects through interfaces and keeping these in a separate module is that it should be possible for implementations to be switched out or changed without affecting applications using the API.

### Module alpha-commons

The commons module holds implementations of the "basic building blocks" of ASP programs, such as terms, atoms, literals, etc. which are used across all other modules. These implementations can be instantiated through factory methods (such as `Atoms.newBasicAtom(...)`), but are not visible to other modules themselves in order to avoid accidental dependencies on implementation-specifc behavior in code using those implementations.

### Module alpha-solver

This module provides an implementation of the `Alpha` interface which is the main API entry point for applications. It is separate from commons and core to avoid API clients having compile-time dependencies against types from `alpha-core`. `alpha-core` is an "implementation" scoped, i.e. "internal" dependency of `alpha-solver`, which means that it is a runtime- but not a compile-time-dependency of code depending on `alpha-solver`.

### Module alpha-core

`alpha-core` implements Alpha's core functionality, i.e. program parsing, program transformations, grounding and solving. It depends on `alpha-commons` and `alpha-api`. While the core module is by far the largest, it is not intended for applications to directly depend on `alpha-core`, but rather use it through the public API realized in the  api, commons, and solver packages.

### Module alpha-cli-app

The cli-app module uses the public API to provide Alpha's functionality in a commandline application. It holds a `Main` class, commandline argument parsing and various utilities for use of Alpha as a commandline-application.


lorenzleutgeb:
Hey @madmike200590! Happy to see that you decided to jump in and continue. To be honest I anticipated the split to go much smoother. Turns out that I forgot that we need to split the API from implementations (which causes quite a number of changes), and there are some rough edges like the "Standard Library".

madmike200590:
> Hey @madmike200590! I am happy to see that you decided to jump in and continue. To be honest I anticipated the split to go much smoother. 

Thank you for getting this off the ground! I've wanted to do something in this branch for some time now, but was tied up with PR #268.

> Turns out that I forgot that we need to split the API from implementations (which causes quite a number of changes), and there are some rough edges like the "Standard Library".

Yeah, I also only fully realized that when looking at your changes. Still not sure about the nicest way to go, but here are my thoughts so far:
* We definitely want a split between "core API" (what the solver uses) and public API, because we don't want methods like `FixedInterpretationLiteral#getSatisfyingSubstitutions` in public API types.
* On the other hand, we have to make sure that everything used in the solver itself conforms to the "core API", ideally that can be statically guaranteed without weird `instanceof` constructs that only blow up at runtime.

So what I'm trying to do now is to create a separate hierarchy of "Core" types, i.e. `CoreLiteral`, `CoreAtom`, etc. that do **not** extend the "public API" types, but are created from them using mapping methods, e.g. `CoreLiteral#fromLiteral` or something like that. While that undoubtedly adds some overhead and very similar code in a few places, it's the neatest way I see to fully decouple the public API from internal workings of our backend and vice-versa. I expect to have a first working draft of that by end of this week, let's see how it looks when it's actually there and doing something...

lorenzleutgeb:
I see. Aggregates are definitely very useful as well!

And yes, I also ran into the same troubles (public API and internal/core API). I wanted to resolve it with generics, but this is very tedious so I gave up. Your approach with mapping the public types to core types is very valid, too, give it a spin!

I can't wait to add Alpha (just the solver, not the CLI) as a Gradle dependency in some application project for the first time 😆

madmike200590:
@lorenzleutgeb, I have a gradle-noob-question that formed in my head while fiddling apart public and "core-internal" APIs...
Ultimately, I think we want to ensure that the `core` module is only ever a runtime dependency for any application using Alpha through the API defined in `api`. Especially, I think we want to ensure that it is not (easily through a "normal build workflow") possible for an API consumer to directly compile against types from `core` (e.g. `InternalProgram` etc).

In order to achieve that, we need to have all relevant functionality exposed through interfaces in the `api` module.
*However*, we also need users to be able to instantiate implementations. Considering the following code snippet from a hypothetical application `api-using-app`:
```
class App {
    public static void main(String[] args){
        Alpha alpha = new AlphaImpl();
        alpha.solve(....);
        ....
    }
}
```
Since `AlphaImpl` resides in `core`, our hypothetical API consumer would need a gradle config looking something like
```
dependencies {
	implementation group: 'at.ac.tuwien.kr',     name: 'alpha-api',      version: '1.0.0'
	implementation group: 'at.ac.tuwien.kr',     name: 'alpha-core',    version: '1.0.0'
}
```
In my opinion, this would be a problem, since it would make expose the types meant for internal use from `alpha-core` while the only type that *should* be exposed is `AlphaImpl`.

The only way around this that I currently see is to have an additional module, let's call it `alpha-solver` that basically just contains `AlphaImpl` and has the following dependencies:
```
dependencies {
	api group: 'at.ac.tuwien.kr',                        name: 'alpha-api',      version: '1.0.0'
	implementation group: 'at.ac.tuwien.kr',    name: 'alpha-core',    version: '1.0.0'
}
```
As far as I understand gradle dependency scopes, that would mean applications depending on `alpha-solver` can (as intended) compile against `alpha-api` as well as `AlphaImpl`, but get `alpha-core` (against which `alpha-solver` is compiled) only as runtime dependency, thereby ensuring they cannot accidentally use types from `alpha-core` in their application code.

Since I'm much more confident with maven than gradle,
- Is this the right way to go about this in gradle?
- Can you think of an alternative that achieves the same without the additional module? I can't think of anything simpler, but still feels like a bit much overhead...

lorenzleutgeb:
Hmm, I am not sure whether Gradle is the correct layer to address this. I guess it should be [JPMS](https://en.wikipedia.org/wiki/Java_Platform_Module_System) which allows us to specify which packages of our module should be visible to the outside. And then being careful about marking classes `pubic`.
I think adding another module or adding JPMS config can also be done rather independently once modularization is done.

lorenzleutgeb:
~350 files changed. That might be ... all of them? 😂

madmike200590:
> ~350 files changed. That might be ... all of them?

pretty much :joy: . Just getting everything to compile took quite some adaptations. Right now we have a lot of interfaces in the API module, a number of which I don't think should actually be exposed, but I couldn't keep everything straight in my head all at once. So now that it compiles, I'm in the process of getting all tests to pass, then cut down on publicly visible interfaces.

AntoniusW:
The `commons` module is confusing me, as it contains some very, very core concepts like the `Interner.java`, which is the implementation ensuring that same-term-is-same-object. This should be in the `core` and not be visible outside.

The `commons` module also contains factory methods for terms/predicates/atoms which looks good to me. But I do not understand why all the api-implementations of those objects reside in the `commons` module. Especially given the fact that all those api-implementations are also in the `commons` module, hence those factories are not much else than constructor-wrappers when the constructors themselves have the same visibility as the factory methods.

I also wonder why construction methods for programs are not part of the `commons` module. Gtiven that terms and atoms are there, I would expect that at least some ways to construct  an `InputProgram` are in `commons`.

lorenzleutgeb:
Okay. I think we're on the same page now. Please bear with me, as I am attending a summer school this week and the schedule is quite tight. I'll find make time for a proper session on 2021-10-19.

madmike200590:
> General question at @madmike200590: have the tests in this PR been converted to use junit 5? 

Yes.

> It seems to me that this branch is still on junit 4 (and if we proceed with a step-wise modularization of master, it maybe won't be necessary to merge junit5 changes into this branch).

The juni4 dependencies that you saw where accidental leftovers that were there because I overlooked something in the build config. I fixed that now, everything runs on junit5 (only).

lorenzleutgeb:
I guess `at.ac.tuwien.kr.alpha.grounder.rete.*` can be deleted?!

madmike200590 (approved):
I've gone through the code, removed some obsolete `TODO`s and improved javadoc a bit. While it could still look a whole lot better, I think we can merge now.

alpha-api/src/main/java/at/ac/tuwien/kr/alpha/api/programs/NormalProgram.java
```diff
@@ -0,0 +1,12 @@
+package at.ac.tuwien.kr.alpha.api.programs;
+
+import at.ac.tuwien.kr.alpha.api.rules.NormalRule;
+
+/**
+ * A {@link Program} consisting only of facts and {@link NormalRule}s, i.e. no disjunctive- or choice-rules.
```
AntoniusW, line 6:
```suggestion
 * A {@link Program} consisting only of facts and {@link NormalRule}s, i.e. no disjunctive- or choice-rules, and no aggregates in rule bodies.
```

alpha-api/src/main/java/at/ac/tuwien/kr/alpha/api/programs/Predicate.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+public interface Predicate extends Comparable<Predicate> {
+
+	/**
+	 * The predicate symbol.
+	 */
+	String getName();
+
+	/**
+	 * The arity, i.e. number of arguments of an instance, of this predicate.
```
AntoniusW, line 16:
```suggestion
	 * The arity, i.e. number of arguments/terms of an instance over this predicate.
```

alpha-api/src/main/java/at/ac/tuwien/kr/alpha/api/programs/Predicate.java
```diff
@@ -0,0 +1,41 @@
...
+
+		if (result != 0) {
+			return result;
+		}
+
+		return Integer.compare(getArity(), other.getArity());
+	}
+
+	/**
+	 * Indicates whether this predicate is internal. Internal predicates are no shown in answer sets.
```
AntoniusW, line 32:
```suggestion
	 * Indicates whether this predicate is internal. Internal predicates are not shown in answer sets.
```

alpha-api/src/main/java/at/ac/tuwien/kr/alpha/api/programs/analysis/DependencyGraph.java
```diff
@@ -0,0 +1,37 @@
...
+
+import at.ac.tuwien.kr.alpha.api.programs.Predicate;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * The predicate-level dependency graph of an ASP program.
+ * Each node of the dependency graph represents one {@link Predicate}.
+ * An edge between from node A to node B indicates that predicate B depends on predicate A. Edges also store a "sign" (true or false)
```
AntoniusW, line 11:
```suggestion
 * An edge from node A to node B indicates that predicate B depends on predicate A. Edges also store a "sign" (true or false)
```

alpha-api/src/main/java/at/ac/tuwien/kr/alpha/api/programs/atoms/AggregateAtom.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.List;
+
+import at.ac.tuwien.kr.alpha.api.ComparisonOperator;
+import at.ac.tuwien.kr.alpha.api.programs.literals.AggregateLiteral;
+import at.ac.tuwien.kr.alpha.api.programs.literals.Literal;
+import at.ac.tuwien.kr.alpha.api.terms.Term;
+import at.ac.tuwien.kr.alpha.api.terms.VariableTerm;
+
+/**
+ * An {@link Atom} representing a comparison over an aggregate function.
```
AntoniusW, line 12:
```suggestion
 * An {@link Atom} representing a comparison over an aggregate function, for example '13 < #sum{ X : p(X) }'.
```

alpha-api/src/main/java/at/ac/tuwien/kr/alpha/api/programs/literals/Literal.java
```diff
@@ -0,0 +1,39 @@
...
+import at.ac.tuwien.kr.alpha.api.terms.Term;
+import at.ac.tuwien.kr.alpha.api.terms.VariableTerm;
+
+/**
+ * A literal according to the ASP Core 2 Standard.
+ * Wraps an {@link Atom} that may or may not be negated.
+ *
+ * Copyright (c) 2017-2021, the Alpha Team.
+ */
+// TODO go through implementations and pull out stuff that can be default-implemented here
```
AntoniusW, line 18: Is this TODO planned to be addressed in this PR?

alpha-api/src/main/java/at/ac/tuwien/kr/alpha/api/terms/ArithmeticTerm.java
```diff
@@ -0,0 +1,18 @@
+package at.ac.tuwien.kr.alpha.api.terms;
+
+/**
+ * A {@link Term} that is the result of an arithmetic calculation.
```
AntoniusW, line 4:
```suggestion
 * A {@link Term} that is a binary arithmetic expression.
```

alpha-api/src/main/java/at/ac/tuwien/kr/alpha/api/programs/literals/Literal.java
madmike200590: No, this should be addressed as part of sub-item "Revisit and potentially refactor Atom and Literal hierarchies" of #300 and should be left in there until such a time.

buildSrc/src/main/kotlin/alpha.java-common-conventions.gradle.kts
```diff
@@ -0,0 +1,67 @@
...
+
+// Fix checkstyle version.
+checkstyle {
+	toolVersion = "7.6"
+}
+
+tasks.withType<JavaCompile> {
+    val compilerArgs = options.compilerArgs
+    //compilerArgs.add("-Xdoclint:all,-missing")
+    compilerArgs.add("-Xlint:all")
```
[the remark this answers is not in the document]
AntoniusW: @lorenzleutgeb I have a regression here: the current settings again make the compiler ignore any `@SuppressWarnings("unchecked")` statements.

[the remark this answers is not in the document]
AntoniusW: Ah, my warning suppression was not correct, it needs to be `@SuppressWarnings(value = {"rawtypes", "unchecked"})` and then the warning no longer appears.

alpha-cli-app/build.gradle.kts
```diff
@@ -0,0 +1,64 @@
...
+
+dependencies {
+    implementation(project(":alpha-solver"))
+    implementation(project(":alpha-commons"))
+
+	implementation("commons-cli:commons-cli:1.3.1")
+
+	val poiVersion = "4.1.1"
+	implementation("org.apache.poi:poi:${poiVersion}")
+	implementation("org.apache.poi:poi-ooxml:${poiVersion}")
```
AntoniusW, line 13: Is the indentation intended to be different in these lines?

alpha-cli-app/build.gradle.kts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+val main = "at.ac.tuwien.kr.alpha.Main"
+
+application {
+    mainClass.set(main)
+}
+
+tasks.create<Jar>("bundledJar") {
```
AntoniusW, line 22: When running the `bundledJar` task using `gradlew` it fails
```
./gradlew bundledJar
> Task :alpha-cli-app:bundledJar FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':alpha-cli-app:bundledJar'.
> Entry at/ac/tuwien/kr/alpha/api/impl/AlphaImpl$1.class is a duplicate but no duplicate handling strategy has been set. Please refer to https://docs.gradle.org/7.3/dsl/org.gradle.api.tasks.Copy.html#org.gradle.api.tasks.Copy:duplicatesStrategy for details.

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 8s
20 actionable tasks: 2 executed, 18 up-to-date
```
It seems that the duplicate originates in some copy operation of the `bundledJar` task.
Is it possible that the `compileClasspath` contains files that are also in the `runtimeClasspath` and hence duplicated?

lorenzleutgeb: This should be fixed in b07a088b. The mistake was duplicating `with(tasks["jar"] as CopySpec)` which would lead to all files being copied twice.

AntoniusW: Thanks, now I can build the bundledJar.

However, a follow-up issue: when I run the bundled jar, it shows me some SLF4J warnings as follows:
```
java -jar alpha-cli-app-bundled.jar -i ASPCOMP2013-instances/b27/instances/0001-graph_colouring-125-0.asp 
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
Answer set 1:
{ ...
```
@lorenzleutgeb any idea what happens here?

alpha-commons/src/main/java/at/ac/tuwien/kr/alpha/commons/PredicateImpl.java
```diff
@@ -1,33 +1,39 @@
...
 		this.name = name;
 		this.arity = arity;
 		this.internal = internal;
 		this.solverInternal = solverInternal;
 	}
 
-	public static Predicate getInstance(String symbol, int arity) {
+	static PredicateImpl getInstance(String symbol, int arity) {
 		return getInstance(symbol, arity, false, false);
 	}
```
AntoniusW, line 29: This method is never used as `Predicates` does its job already.
```suggestion
```

alpha-commons/src/main/java/at/ac/tuwien/kr/alpha/commons/PredicateImpl.java
```diff
@@ -1,33 +1,39 @@
...
 
-	public static Predicate getInstance(String symbol, int arity, boolean internal) {
+	static PredicateImpl getInstance(String symbol, int arity, boolean internal) {
 		return getInstance(symbol, arity, internal, false);
 	}
```
AntoniusW, line 33: This method is never used as `Predicates` does its job already.
```suggestion
```

alpha-cli-app/build.gradle.kts
lorenzleutgeb: Yes, we do not have any implementation of a logger in the runtime classpath of `alpha-cli-app`. I'll add one.

lorenzleutgeb: Fixed in 073da170.

alpha-commons/src/main/java/at/ac/tuwien/kr/alpha/commons/atoms/ExternalAtomImpl.java
```diff
@@ -1,19 +1,19 @@
 /**
- * Copyright (c) 2017-2020, the Alpha Team.
+ * Copyright (c) 2017-2019, the Alpha Team.
```
AntoniusW, line 2:
```suggestion
 * Copyright (c) 2017-2021, the Alpha Team.
```

alpha-commons/src/main/java/at/ac/tuwien/kr/alpha/commons/literals/BasicLiteralImpl.java
```diff
@@ -25,51 +25,55 @@
...
- * Contains a potentially negated {@link BasicAtom}.
- * 
+ * Contains a potentially negated {@link BasicAtomImpl}.
+ *
  * Copyright (c) 2017-2018, the Alpha Team.
  */
-public class BasicLiteral extends Literal {
-	
-	public BasicLiteral(BasicAtom atom, boolean positive) {
+class BasicLiteralImpl extends AbstractLiteral implements BasicLiteral { // TODO could we parameterize Literal with Atom type?
```
AntoniusW, line 47: The TODO is a discussion item for an issue, not the actual code here.
```suggestion
class BasicLiteralImpl extends AbstractLiteral implements BasicLiteral {
```

alpha-cli-app/build.gradle.kts
```diff
@@ -0,0 +1,73 @@
...
+
+	archiveFileName.set("${project.name}-bundled.jar")
+
+	exclude("META-INF/DEPENDENCIES")
+	
+	/*
+	 * In order to make sure we don"t overwrite NOTICE and LICENSE files coming from dependency
+	 * jars with each other, number them while copying
+	 */
+	var noticeCount = 1
```
AntoniusW, line 48: Currently renaming starts at `2`.
```suggestion
	var noticeCount = 0
```

alpha-cli-app/build.gradle.kts
```diff
@@ -0,0 +1,73 @@
...
+	rename { it : String ->
+		return@rename if ("NOTICE.txt".equals(it) || "NOTICE".equals(it)) {
+			noticeCount++
+			"NOTICE.${noticeCount}.txt"
+		} else {
+			it
+		}
+	}
+
+	var licenseCount = 1
```
AntoniusW, line 58: Currently renaming starts at `2`.
```suggestion
	var licenseCount = 0
```

alpha-commons/src/main/java/at/ac/tuwien/kr/alpha/commons/comparisons/ComparisonOperators.java
```diff
@@ -0,0 +1,95 @@
...
+			return t1.compareTo(t2) <= 0;
+		}
+
+	};
+
+	public static final ComparisonOperator GE = new AbstractComparisonOperator(">=") {
+
+		@Override
+		public ComparisonOperator negate() {
+			return ComparisonOperators.LE;
```
AntoniusW, line 81: This looks very wrong!
```suggestion
			return ComparisonOperators.LT;
```

alpha-core/build.gradle.kts
```diff
@@ -0,0 +1,55 @@
...
+
+	// NOTE: Contents of the antlr subpackage are autogenerated (see configuration of
+	//	   AntlrTasks above). It does not make sense to include them in our coverage
+	//	   report.
+	// TODO: Translate to Kotlin.
+	//afterEvaluate {
+		//getClassDirectories().setFrom(files(classDirectories.files.collect {
+			//fileTree(dir: it, exclude: "at/ac/tuwien/kr/alpha/antlr/**")
+		//}))
+	//}
```
AntoniusW, line 51: Is this TODO still relevant?

lorenzleutgeb: Fixed in 33b8c121

alpha-commons/src/main/java/at/ac/tuwien/kr/alpha/commons/comparisons/ComparisonOperators.java
lorenzleutgeb: Fixed in 77474e9

alpha-core/src/main/java/at/ac/tuwien/kr/alpha/core/programs/transformation/NormalizeProgramTransformation.java
```diff
@@ -21,21 +21,21 @@ public NormalizeProgramTransformation(AggregateRewritingConfig aggregateCfg) {
...
 		tmpPrg = new AggregateRewriting(aggregateRewritingCfg.isUseSortingGridEncoding(), aggregateRewritingCfg.isSupportNegativeValuesInSums()).apply(tmpPrg);
 		// Transform enumeration atoms.
 		tmpPrg = new EnumerationRewriting().apply(tmpPrg);
 		EnumerationAtom.resetEnumerations();
 
 		// Construct the normal program.
-		NormalProgram retVal = NormalProgram.fromInputProgram(tmpPrg);
-		// Transform intervals.
+		NormalProgram retVal = NormalProgramImpl.fromInputProgram(tmpPrg);
+		// Transform intervals
```
AntoniusW, line 38:
```suggestion
		// Transform intervals.
```

AntoniusW (changes requested):
Adding this as intermediate review: some comments need addressing. Currently at 307/436 reviewed files and progressing. We are getting there. :-)

alpha-core/src/main/java/at/ac/tuwien/kr/alpha/core/solver/DefaultSolver.java
```diff
@@ -61,23 +46,38 @@
...
+import at.ac.tuwien.kr.alpha.core.solver.learning.GroundConflictNoGoodLearner;
+import at.ac.tuwien.kr.alpha.core.util.Substitutions;
 
 /**
- * The default solver employed in Alpha.
- * The core algorithm is a DPLL-style loop employing conflict-driven learning and lazy-grounding of nogoods.
+ * The new default solver employed in Alpha.
  *
- * Copyright (c) 2016-2021, the Alpha Team.
+ * Copyright (c) 2016-2020, the Alpha Team.
```
AntoniusW, line 78:
```suggestion
 * Copyright (c) 2016-2021, the Alpha Team.
```

alpha-core/src/test/java/at/ac/tuwien/kr/alpha/core/test/util/RuleParser.java
```diff
@@ -0,0 +1,23 @@
...
+import at.ac.tuwien.kr.alpha.api.rules.heads.Head;
+import at.ac.tuwien.kr.alpha.core.parser.ProgramParserImpl;
+
+public class RuleParser {
+
+	public static Rule<Head> parse(String str) {
+		ProgramParser parser = new ProgramParserImpl();
+		ASPCore2Program prog = parser.parse(str);
+		if (!prog.getFacts().isEmpty()) {
+			throw new IllegalArgumentException("Excepted exactly one rule and no facts!");
```
AntoniusW, line 15:
```suggestion
			throw new IllegalArgumentException("Expected exactly one rule and no facts!");
```

alpha-core/src/test/java/at/ac/tuwien/kr/alpha/core/test/util/RuleParser.java
```diff
@@ -0,0 +1,23 @@
...
+		}
+		if (prog.getRules().size() != 1) {
+			throw new IllegalArgumentException("Excepted exactly one rule");
```
AntoniusW, line 18:
```suggestion
			throw new IllegalArgumentException("Expected exactly one rule");
```

AntoniusW:
Okay, I am done with the reviewing, all files viewed, finally. There are some of my comments in the last week that should be addressed and there is a git conflict with the `build.gradle`. But besides that, this PR is ready for merging! Very good work, thank you @lorenzleutgeb and @madmike200590 !

lorenzleutgeb:
Regarding the conflicting file `build.gradle`. The conflict is as follows:

```
CONFLICT (modify/delete): build.gradle deleted in modules and modified in master.
                          Version master of build.gradle left in tree.
```
This is just because we have changes to `build.gradle` on `master`, and deleted it in `modules`. Funnily enough, GitHub opines that "These conflicts are too complex to resolve in the web editor".

I merged `master` into `modules` ("control merge") to resolve, resulting in c65b8ea67420cf8e0a2d862264a48e39b44c3796.

alpha-commons/src/main/java/at/ac/tuwien/kr/alpha/commons/literals/BasicLiteralImpl.java
lorenzleutgeb: Mentioned in #300

alpha-commons/src/main/java/at/ac/tuwien/kr/alpha/commons/PredicateImpl.java
lorenzleutgeb: Deleted in d7f71891

alpha-commons/src/main/java/at/ac/tuwien/kr/alpha/commons/PredicateImpl.java
lorenzleutgeb: Deleted in d7f71891

alpha-core/src/main/java/at/ac/tuwien/kr/alpha/core/common/fixedinterpretations/PredicateInterpretationImpl.java
```diff
@@ -0,0 +1,13 @@
+package at.ac.tuwien.kr.alpha.core.common.fixedinterpretations;
+
+import java.util.List;
+import java.util.Set;
+
+import at.ac.tuwien.kr.alpha.api.common.fixedinterpretations.PredicateInterpretation;
+import at.ac.tuwien.kr.alpha.api.terms.ConstantTerm;
+import at.ac.tuwien.kr.alpha.api.terms.Term;
+
+public interface PredicateInterpretationImpl extends PredicateInterpretation {
```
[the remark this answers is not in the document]
lorenzleutgeb: Deleted in d7f71891

alpha-core/src/main/java/at/ac/tuwien/kr/alpha/core/common/CoreAnswerSet.java
```diff
@@ -0,0 +1,14 @@
+package at.ac.tuwien.kr.alpha.core.common;
+
+import java.util.SortedSet;
+
+import at.ac.tuwien.kr.alpha.api.programs.Predicate;
+import at.ac.tuwien.kr.alpha.api.programs.atoms.Atom;
+
+public interface CoreAnswerSet extends Comparable<CoreAnswerSet> {
```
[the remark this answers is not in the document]
lorenzleutgeb: Deleted in d7f71891

alpha-cli-app/build.gradle.kts
AntoniusW: @lorenzleutgeb not sure what happened to this item here. Is there a reason we want the licenses to start from `LICENSE.2.txt` and not `LICENSE.1.txt`?

AntoniusW (approved):
I would like to have a response for my old comment in `alpha-cli-app/build.gradle.kts` but other than that, everything is fine!

alpha-cli-app/build.gradle.kts
lorenzleutgeb: Sorry, I missed that this discussion is still unresolved.

Fixed in bddb5d60.

lorenzleutgeb:
Looks like we're good to merge, yay!

@madmike200590 do you want to rebase this branch, to clean up the history? There are commits messages starting with "WIP" or "Intermediate commit". I'd like to avoid such commits on `master`. At the same time, I'd also like to avoid squashing this rather large branch into one single commit.

madmike200590:
> Looks like we're good to merge, yay!
> 
> @madmike200590 do you want to rebase this branch, to clean up the history? There are commits messages starting with "WIP" or "Intermediate commit". I'd like to avoid such commits on `master`. At the same time, I'd also like to avoid squashing this rather large branch into one single commit.

Great! While I see your point about the WIP commits, I don't really want to do a rebase only for that - especially since it's not like it'd be useful to try checking out "halfway done" versions of the modularization. If that's ok with everyone, I'd just go ahead and squash everything.

lorenzleutgeb:
I took the liberty to squash and merge, cutting down the merge commit message *a lot*. I used [`Co-authored-by`](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) ([also used in the Linux kernel](https://git.wiki.kernel.org/index.php/CommitMessageConventions)) to indicate that we share authorship of the commit. I'm really happy that we made it :)

madmike200590:
Great, thanks! It's a huge step forward and after all the time this took, it's great to have this finally "live" :D

PR #279, opened 2020-12-04
Fix accumulator flag in grounder being ignored

### LINKED ISSUE #271: Grounder accumulator flag ignored
Apparently, the grounder refactoring made it ignore the accumulator flag as `GrounderHeuristicsConfiguration::isAccumulatorEnabled()` is never used. If the `-acc` flag is enabled many instances are still removed from `IndexedInstanceStorage`, which the accumulator should rather not do.

madmike200590:
Closes #271 

During the refactoring of the grounder performed in #256, the handling for the accumulator flag in NaiveGrounder was overlooked.
That logic has now been added to `DefaultLazyGroundingInstantiationStrategy`, such that `Atom`s that exist in the working memory are always reported as having `AssignmentStatus#TRUE`.

@AntoniusW I believe that this should be fixed with my change, but I find it hard to devise a test case that would properly test correct functioning of the accumulator strategy. Do you have some code handy for that, or could you give me some pointers on how to set up that scenario and what/when to check exactly?


AntoniusW:
> @AntoniusW I believe that this should be fixed with my change, but I find it hard to devise a test case that would properly test correct functioning of the accumulator strategy. Do you have some code handy for that, or could you give me some pointers on how to set up that scenario and what/when to check exactly?

I think the following ASP program would allow to test the accumulator property:
```
{ a } .
{ b } :- a.
{ c } :- not a.
p(3) :- b.
q(3) :- c.
r(X) :- p(X), q(X).
```
After computing all answer-sets for this program, the `AtomStore` should know an atom `r(3)` and there should be a NoGood with 3 literals representing the rule body of the last rule with `X=3` which the grounder should have handed-over to the solver. Now, one either has to check this property using the DefaultSolver, or one simulates its behaviour in the unit test. For the former, I think it is a bit harder to check, since the `DefaultSolver` does not allow inspection of its known nogoods. Maybe a custom grounder, that holds a `NaiveGrounder` internally and forwards all calls to that `NaiveGrounder` but also records grounded nogoods returned to the solver, might be a good way to see whether the above nogood is grounded. Simulating the behaviour of the `DefaultSolver` on the other hand is probably more cumbersome since there is quite a number of rounds between solver and grounder and each round needs a correct assignment.

AntoniusW (approved):

PR #282, opened 2020-12-15
Additional constraint generated during transformation of choice rules

rtaupe:
In addition to the two normal rules encoding a choice rule, `ChoiceHeadToNormal` now also generates a constraint that forbids the two complementary head literals to be true at the same time. The constraint is grounded at the same time as the choice rule itself, helps propagation and thus potentially improves solving performance.

### DIFF
--- a/src/main/java/at/ac/tuwien/kr/alpha/grounder/transformation/ChoiceHeadToNormal.java
+++ b/src/main/java/at/ac/tuwien/kr/alpha/grounder/transformation/ChoiceHeadToNormal.java
@@ -1,5 +1,5 @@
-/**
- * Copyright (c) 2017-2018, the Alpha Team.
+/*
+ * Copyright (c) 2017-2020, the Alpha Team.
  * All rights reserved.
  *
  * Additional changes made by Siemens.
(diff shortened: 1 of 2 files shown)


rtaupe:
I´ve marked this pull request as draft because I still want to do some performance tests.

rtaupe:
This feature does not seem to achieve the expected performance benefits (at least in some cases).
For example, a [pup instance](https://github.com/alpha-asp/Alpha/files/5702928/pup.zip) is solved within ~11 seconds (15k backjumps) without this feature and within ~30s (44k backjumps) with this feature.

The alternative approach #284 to solve this issue, however, shows benefits on this instance: ~7s (5k backjumps).

Maybe further performance analysis is in order before either of the two approaches is fully implemented and merged?

AntoniusW:
What is the status on this PR and #284 ?

From my current understanding, it feels a bit like the additional nogood/constraint might be superfluous. At least in cases where completion nogoods are added (i.e., currently only if the head contains all variables of the body and no other rule can derive the head predicate), the added constraint is already implied via the completion.

rtaupe:
> 
> 
> What is the status on this PR and #284 ?
> 
The status has not changed since Dec 16, 2020, I have not done any additional performance analysis since then.

Issue #193, opened 2019-07-12
Javadoc errors


rtaupe:
Generating javadoc using `gradle javadoc` yields many warnings and even errors, for example:

```
...\alpha-asp\Alpha\src\main\java\at\ac\tuwien\kr\alpha\grounder\Unifier.java:83: error: bad use of '>'
         * if left has A -> B and right has A -> t then the result will have A -> t and B -> t.
```

PR #290, opened 2021-06-17
Fix javadoc error in Unifier

kharus:
javadoc doesn't like `>` symbols. One way to solve this is to wrap them
`@literal`
#193


lorenzleutgeb (approved):

lorenzleutgeb:
IMO it might be even better to use `&rarr;` or a [unicode arrow](https://unicode.org/charts/nameslist/n_2190.html), such as `→`. But that's just a side note.

kharus:
Actually Unicode arrow is a good idea as it looks the same in java sources (as oppose to `&rarr;`)  and resulting html.

PR #296, opened 2021-08-29
ci: Migrate to GitHub Actions

### LINKED ISSUE #205: Migrate to GitHub Actions
Use [GitHub Actions](https://github.com/features/actions) instead of Travis CI and AppVeyor.

lorenzleutgeb:
Created `.github/workflows/build.yml` to supersede `.travis.yml` and `.appveyor.yml`. This will need some tweaking...

FYI as of 2021-08-29 I suspended the "Travis CI" GitHub App for the "alpha-asp" organisation and disabled outgoing webhooks from GitHub to AppVeyor.

Closes #205


lorenzleutgeb:
As of 528b23f we have a working build matrix. We do not yet extract artifacts (like built JARs) but I would like to ask for review and possibly merge anyways so that we at least have a CI system in place again. I can add artifact upload (and more) later.

AntoniusW:
Hi @lorenzleutgeb,
first of all, thank you for your continued support! I am currently trying to understand what changes and to what effects. I think it is a good decision to get rid of TravisCI, which showed questionable behaviour recently (including security issues). At the moment, I do not yet fully understand what exactly happens here with GitHub Actions, so I have no review yet, but still some questions:

1. To my understanding, we get 2000 minutes of runtime per month from GitHub Actions, right? At the moment, our intensive CI tests run for about one hour and the matrix here tests that on 9 combinations of java x operating system. So that means we have about 4 full CI runs (triggered by pushes into the repo) per month. Is that correct?

2. If I understand this PR correctly, then it also disables the AppVeyor CI. Uhh, why?

3. The build contains a `jacocoTestReport coveralls` and this script apparently failes (see: https://github.com/alpha-asp/Alpha/runs/3454715203?check_suite_focus=true ). Did we have this enabled previously, or is that new?

Thanks again!

lorenzleutgeb:
> Hi @lorenzleutgeb,
> first of all, thank you for your continued support! I am currently trying to understand what changes and to what effects. I think it is a good decision to get rid of TravisCI, which showed questionable behaviour recently (including security issues). At the moment, I do not yet fully understand what exactly happens here with GitHub Actions, so I have no review yet, but still some questions:

Sure :)

> 1. To my understanding, we get 2000 minutes of runtime per month from GitHub Actions, right? At the moment, our intensive CI tests run for about one hour and the matrix here tests that on 9 combinations of java x operating system. So that means we have about 4 full CI runs (triggered by pushes into the repo) per month. Is that correct?

I fail to verify the quota of 2000min p.m.
I read https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration#usage-limits but could not find such quota. Could you please state your source?

> 2. If I understand this PR correctly, then it also disables the AppVeyor CI. Uhh, why?

The original reason for using AppVeyor was that we wanted to have some CI that builds and tests on Windows and Travis CI does not offer running jobs on Windows. However, GitHub Actions does offer running jobs on Windows. By just using GitHub Actions we can decrease complexity of our configuration (now just one instead of two CI services) and still build and test on Windows. If there's any particular reason to keep AppVeyor, we can do so, of course. I just don't see any.

> 3. The build contains a `jacocoTestReport coveralls` and this script apparently failes (see: https://github.com/alpha-asp/Alpha/runs/3454715203?check_suite_focus=true ). Did we have this enabled previously, or is that new?

Yes, we did run this before, see https://github.com/alpha-asp/Alpha/blob/b1fc76c89b40e8305e6a999b38c1d309d7aa93d8/.travis.yml#L54. I'll look into it and will have a more detailed response next week.

AntoniusW:
> > 1. To my understanding, we get 2000 minutes of runtime per month from GitHub Actions, right? At the moment, our intensive CI tests run for about one hour and the matrix here tests that on 9 combinations of java x operating system. So that means we have about 4 full CI runs (triggered by pushes into the repo) per month. Is that correct?
> 
> I fail to verify the quota of 2000min p.m.
> I read https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration#usage-limits but could not find such quota. Could you please state your source?
> 
You are right. I saw that for GitHub Actions there are 2000 minutes per month for organizations in the Free plan (which we have), but I didn't see that GitHub Actions usage is free for public repositories (as stated in https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions).
> > 1. If I understand this PR correctly, then it also disables the AppVeyor CI. Uhh, why?
> 
> The original reason for using AppVeyor was that we wanted to have some CI that builds and tests on Windows and Travis CI does not offer running jobs on Windows. However, GitHub Actions does offer running jobs on Windows. By just using GitHub Actions we can decrease complexity of our configuration (now just one instead of two CI services) and still build and test on Windows. If there's any particular reason to keep AppVeyor, we can do so, of course. I just don't see any.
> 
That is a good argument. I agree.

> > 1. The build contains a `jacocoTestReport coveralls` and this script apparently failes (see: https://github.com/alpha-asp/Alpha/runs/3454715203?check_suite_focus=true ). Did we have this enabled previously, or is that new?
> 
> Yes, we did run this before, see
> 
> https://github.com/alpha-asp/Alpha/blob/b1fc76c89b40e8305e6a999b38c1d309d7aa93d8/.travis.yml#L54
> . I'll look into it and will have a more detailed response next week.

Okay, I must have ignored it for most of the time (and it is now down since a few days).

Thanks for your responses, @lorenzleutgeb

.github/workflows/build.yml
```diff
@@ -0,0 +1,73 @@
+name: Build
+on: [push]
```
AntoniusW, line 2: Maybe this list should include `pull_request`?

.github/workflows/build.yml
```diff
@@ -0,0 +1,73 @@
...
+    - uses: actions/checkout@v2
+      with:
+        submodules: recursive
+        fetch-depth: 0
+    - uses: actions/setup-java@v2
+      with:
+        java-version: ${{ matrix.java }}
+        distribution: 'adopt'
+        architecture: x64
+    - uses: actions/cache@preview
```
AntoniusW, line 57: If I understand actions correctly, then the `@preview` version of the cache action is from 2019 and very outdated by now. Why not use the current `v2` version?

.github/workflows/build.yml
```diff
@@ -0,0 +1,73 @@
...
+      MAIN_OS: ubuntu
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: recursive
+        fetch-depth: 0
+    - uses: actions/setup-java@v2
+      with:
+        java-version: ${{ matrix.java }}
+        distribution: 'adopt'
```
AntoniusW, line 55: It seems that the `adopt` distribution is obsolete now: https://github.com/actions/setup-java/blob/main/docs/advanced-usage.md#adopt
According to that, one should use `temurin` now.

AntoniusW (approved):
Looks good. I still have some questions for certain settings, but aside from that would be ready to merge.

lorenzleutgeb:
Thank you for those helpful remarks, @AntoniusW! You caught me pasting parts of this configuration from another project with outdated parameters :) I addressed all your remarks.

Note that the patch coverage is very low, because we don't have coverage data for our CI configuration.

PR #297, opened 2021-09-27
Refactor DefaultSolver

AntoniusW:
Some refactoring/polishing inside the solver component, mostly DefaultSolver. Changes are intended to support subclassing of DefaultSolver for optimization tasks/weak constraints.

### DIFF
--- a/src/main/java/at/ac/tuwien/kr/alpha/solver/DefaultSolver.java
+++ b/src/main/java/at/ac/tuwien/kr/alpha/solver/DefaultSolver.java
@@ -96,7 +96,6 @@ public class DefaultSolver extends AbstractSolver implements SolverMaintainingSt
 	private static class SearchState {
 		boolean hasBeenInitialized;
 		boolean isSearchSpaceCompletelyExplored;
-		boolean didChange;
 		boolean afterAllAtomsAssigned;		// True if search reached fixpoint and all remaining unassigned atoms have been set to false.
 	}
 	private final SearchState searchState = new SearchState();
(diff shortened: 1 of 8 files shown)


src/main/java/at/ac/tuwien/kr/alpha/solver/DefaultSolver.java
```diff
@@ -124,108 +127,124 @@ private BranchingHeuristic chainFallbackHeuristic(Grounder grounder, WritableAss
...
+	private void setupLoop() {
+		if (!searchState.hasBeenInitialized) {
+			// Initially, get NoGoods from grounder.
+			performanceLog.initialize();
+			getNoGoodsFromGrounderAndIngest();
+			searchState.hasBeenInitialized = true;
+			return;
+		}
+		// We already found one Answer-Set and are requested to find another one.
+		searchState.afterAllAtomsAssigned = false;
```
madmike200590, line 169: If I read the code correctly, we reach this part of the code on the second (and subsequent) call(s) to tryAdvance (as in the first call, we'll return after initializing search space). This is not really intuitive from the comment here, I think readability would benefit from a somewhat more detailed explanation on _why_ we only get there after finding an answer set in the comment.

madmike200590 (approved):
This looks really good! The additional methods in the solver improve readability quite a bit. There's one minor comment where I think a comment could be a bit clearer, but as far as I'm concerned, this is ready to be merged.

src/main/java/at/ac/tuwien/kr/alpha/solver/DefaultSolver.java
AntoniusW: Yes, the code is doing that. I split up the method into two separate ones.

AntoniusW:
@lorenzleutgeb and/or @madmike200590 please have another look.

madmike200590 (approved):
Looks good to me! I think the newly split version of the solver initialization part enhances readability a lot.

PR #298, opened 2021-09-29
Add coveralls.io badge

AntoniusW:
Adds the coveralls.io badge to the repo page.


lorenzleutgeb (approved):

lorenzleutgeb:
I'm okay with merging. At the risk of going off topic: This would add the second badge related to coverage (codecov is already there, this MR adds coveralls). Maybe we should focus on one of the two? Personally I have a slight tendency for coveralls, because I like the comments with coverage diffs that it generates.

AntoniusW:
@lorenzleutgeb Well, our CI build apparently feeds two coverage services, so I think it might be useful to have direct links to both. Other reason for adding is that I was simply ignoring the second one (i.e., coveralls) until now, because it was not present anywhere.

PR #308, opened 2021-11-05
Refactor tests on modularized code structure

madmike200590:
This PR is a follow-up to PR #274. Following the changes to the overall code structure and API introduced through modularization, I took the opportunity to get started on some refactorings of Alphas unit tests.
Specifically, this PR aims to

##### Reduce dependencies and narrow down tested code for unit tests
In order to make our unit tests more precise and avoid localized errors failing lots of tests, I removed unnecessary dependencies from most tests. Especially, we don't want to have tests of core components like gronder and solver depend on the complete program preprocessing workflow including parser and rewritings.

##### Move functional tests to alpha-solver module
Most tests that test rewriting and preprocessing features, as well as all tests that just pass programs into an `Alpha` instance and verify returned answer sets should be considered functional tests rather than unit tests since they test complete features through the complete workflow. These were moved into the `alpha-solver` module since the default implementation of `Alpha` resides there.

##### Mark functional and integration tests for further attention
In order to improve build performance and better distinguish between unit-, integration-, functional- and performance-tests, these should be separated into separate tasks in the build (of which some should probably only run in CI) process.
This has not yet been realized yet, but tests that should probably be treated as integration tests or functional tests were marked with appropriate `TODO`s in preparation for that refactoring step. (Issue TBD)


madmike200590:
Closing this PR due to broken history - will open new PR

PR #310, opened 2021-11-10
gradle: 7.0 -> 7.3

lorenzleutgeb:
Brings us one step closer to #301

PR #312, opened 2021-11-11
gradle: 7.0 -> 7.3

lorenzleutgeb:
This is a port of #310.

We get a new warning:

```
'compileJava' task (current target is 11) and 'compileKotlin' task (current target is 1.8) jvm target compatibility should be set to the same Java version.
```

Looks like https://github.com/gradle/gradle/issues/18935 to me, let's wait and see.


madmike200590 (approved):
Looks good, thanks for the port!

PR #313, opened 2021-11-15
ci: Clean up GitHub Actions spec

lorenzleutgeb:
Improvements to our GitHub Actions configuration:

### Action `gradle/wrapper-validation-action`

Checks that the file at `/gradle/wrapper/gradle-wrapper.jar` is known (on some allowlist). Raises the bar for someone to push a malicious JAR file to this repo.

### Action `gradle/gradle-build-action`

Is a platform-independent way to invoke Gradle, and automatically handles caching. This way we don't have to provide different ways to invoke Gradle based on `matrix.os`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,25 +32,27 @@ jobs:
       # OS to be the "main" one.
       MAIN_OS: ubuntu
     steps:
+    # See https://github.com/marketplace/actions/checkout
     - uses: actions/checkout@v2
       with:
         submodules: recursive
         fetch-depth: 0
+    # See https://github.com/marketplace/actions/gradle-wrapper-validation
+    - uses: gradle/wrapper-validation-action@v1
+    # See https://github.com/marketplace/actions/setup-java-jdk
     - uses: actions/setup-java@v2
       with:
         java-version: ${{ matrix.java }}
         distribution: 'temurin'
-        architecture: x64
-        cache: 'gradle'
-
-    - run: ./gradlew build --stacktrace
-      if: runner.os != 'Windows'
-    - run: .\gradlew build --stacktrace
-      if: runner.os == 'Windows'
-
-    - run: ./gradlew jacocoTestReport coveralls
+    # See https://github.com/marketplace/actions/gradle-build-action
+    - uses: gradle/gradle-build-action@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -32,25 +32,27 @@ jobs:
...
-      if: runner.os == 'Windows'
-
-    - run: ./gradlew jacocoTestReport coveralls
+    # See https://github.com/marketplace/actions/gradle-build-action
+    - uses: gradle/gradle-build-action@v2
+      with:
+        arguments: build --stacktrace
+    - uses: gradle/gradle-build-action@v2
       if: matrix.os == env.MAIN_OS && matrix.java == env.MAIN_JAVA
       continue-on-error: true
```
AntoniusW, line 53: Question regarding the `continue-on-error` in connection with the `gradle/wrapper-validation-action`: What happens if the validation fails for the main-os/java build? Will it continue or will it stop?

lorenzleutgeb: I am not sure I understand your question. `continue-on-error` is an attribute on the step that uses `gradle/gradle-build-action` (line 51). It does not influence what happens in case the step that uses `gradle/wrapper-validation-action` (line 41) fails.

### Failure of `gradle/wrapper-validation-action` (line 41)

We want the job to fail if `gradle/wrapper-validation-action` fails (which it does, because it is *not* have the attribute `continue-on-error: true`). See also [this test](https://github.com/alpha-asp/Alpha/actions/runs/1465628883), where I just did

```
echo "This should error." > gradle/wrapper/gradle-wrapper.jar
```

and indeed the job fails.

### Failure of `gradle/gradle-build-action` (line 51)

This step generates coverage data. It currently has the attribute `continue-on-error: true` because, we want the job to succeed even if it failed to produce/upload coverage data. Of course, coverage data is nice to have and somewhat valuable, and do want to have it, but it's just not critical in terms of the job suceeding/failing. If you think it should fail the job, I'm happy to remove the attribute.

AntoniusW: Many thanks for the clarification. It was not clear to me whether the `continue-on-error` only applies to that one action or the whole job.

AntoniusW (approved):

PR #314, opened 2021-11-18
Add `CITATION.cff`

lorenzleutgeb:
For more context, refer to:
 * [citation-file-format.github.io](https://citation-file-format.github.io/)
 * [Supported by GitHub](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files)
 * [Supported by Zenodo](https://twitter.com/ZENODO_ORG/status/1420357001490706442)
 * [Supported by Zotero](https://twitter.com/zotero/status/1420515377390530560)

The file should not see many changes (just like `AUTHORS` and `CONTRIBUTORS`). The [spec is quite broad and flexible](https://github.com/citation-file-format/citation-file-format/blob/main/schema-guide.md), so we could encode *much* more, but I'd prefer to start with something simple.

Opinions?

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -10,7 +10,7 @@ authors:
 - family-names: "Langowski"
   given-names: "Michael"
   affiliation: "TU Wien"
-  #orcid: TODO
+  orcid: "https://orcid.org/0000-0001-6572-0357"
 - family-names: "Leutgeb"
   given-names: "Lorenz"
   affiliation: "TU Wien"
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,35 @@
+# https://citation-file-format.github.io/
+cff-version: 1.2.0
+message: "If you use this software, please cite it using these metadata."
+title: "Alpha"
+abstract: "A lazy-grounding Answer-Set Programming System"
+# Authors are sorted:
+# First people, sorted by `family-names`, `given-names`.
+# Second entities, sorted by `name`.
+authors:
+- family-names: "Langowski"
+  given-names: "Michael"
+  affiliation: "TU Wien"
+  #orcid: TODO
+- family-names: "Leutgeb"
+  given-names: "Lorenz"
+  affiliation: "TU Wien"
+  orcid: "https://orcid.org/0000-0003-0391-3430"
+- family-names: "Taupe"
+  given-names: "Richard"
+  affiliation: "Siemens AG"
+  orcid: "https://orcid.org/0000-0001-7639-1616"
+- family-names: "Weinzierl"
+  given-names: "Antonius"
+  affiliation: "TU Wien"
+  orcid: "https://orcid.org/0000-0003-2040-6123"
+license: "BSD-2-Clause"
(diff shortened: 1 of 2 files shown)


CITATION.cff
```diff
@@ -0,0 +1,35 @@
...
+  orcid: "https://orcid.org/0000-0003-0391-3430"
+- family-names: "Taupe"
+  given-names: "Richard"
+  affiliation: "Siemens AG"
+  orcid: "https://orcid.org/0000-0001-7639-1616"
+- family-names: "Weinzierl"
+  given-names: "Antonius"
+  affiliation: "TU Wien"
+  orcid: "https://orcid.org/0000-0003-2040-6123"
+- name: "Siemens AG"
```
lorenzleutgeb, line 25: Not sure if Siemens AG as an entity should be listed as an author, or whether it's enough to list it as the affiliation of an author. But I guess it does not hurt.

CITATION.cff
```diff
@@ -0,0 +1,35 @@
...
+message: "If you use this software, please cite it using these metadata."
+title: "Alpha: A lazy-grounding Answer-Set Programming system"
+# Authors are sorted:
+# First people, sorted by `family-names`, `given-names`.
+# Second entities, sorted by `name`.
+authors:
+- family-names: "Langowski"
+  given-names: "Michael"
+  affiliation: "TU Wien"
+  #orcid: TODO
```
lorenzleutgeb, line 12: @madmike200590 Do you have an ORCID?

rtaupe (approved):

CITATION.cff
AntoniusW: IANAL, but my understanding is that Siemens AG is an author in the sense of copyright laws, since they did pay people to work on Alpha (and at the same time they decided that the result of that work is licensed with the BSD 2-Clause license). So the question whether it belongs here depends on what the authors list should contain: the joint copyright holders, or only persons that wrote the code.

lorenzleutgeb: There's no strict definition. I'd guess the latter.

lorenzleutgeb:
Removed the entry for "Siemens AG" from "contributors". It's still there as the "affiliation" of @rtaupe, like it would appear on a paper. Hope that's okay.

AntoniusW:
To my understanding, Siemens AG is an author (in the copyright sense) because they paid someone (@rtaupe specifically) to work on Alpha during their working hours. Siemens AG agreed to the open source BSD license of that work, but they are still an author as far as I understand. I am not a lawyer, but I think we should have Siemens AG in the list of authors.

rtaupe (approved):

CITATION.cff
madmike200590: Since it probably can't hurt, I got one  now - added it in [70c33f2](https://github.com/alpha-asp/Alpha/pull/314/commits/70c33f2644a44d45ae00588ad5c48f2382ca498c).

madmike200590 (approved):

Issue #301, opened 2021-10-19
Upgrade Java Version


lorenzleutgeb:
Migrate Alpha to [Java 17](https://openjdk.java.net/projects/jdk/17/).

[GraalVM 21.3](https://www.graalvm.org/release-notes/21_3/) was released on 2021-10-19 and supports Java 17.

Blockers:
 - #274 
 - https://github.com/gradle/gradle/issues/16857

lorenzleutgeb:
Gradle 7.3 with Support for Java 17 is out: https://docs.gradle.org/7.3/release-notes.html

lorenzleutgeb:
When migrating, make sure that Checkstyle behaves as expected early on.

lorenzleutgeb:
Now that #274 is merged, this is unblocked.

@AntoniusW how do you feel about raising the minimum required JDK version to 17? It's [available for Ubuntu](https://packages.ubuntu.com/search?suite=default&section=all&arch=amd64&keywords=openjdk-17-jdk&searchon=names) in all current versions ([see Ubuntu version history on Wikipedia](https://en.wikipedia.org/wiki/Ubuntu_version_history#Version_timeline)), and since it is the current Java LTS release (see [Java version history on Wikipedia](https://en.wikipedia.org/wiki/Java_version_history)) it will be supported until 2029. Alpha skipped one LTS release, namely Java 11, in between.

Changing the minimum required JDK version to Java 17, does have some implications regarding dependencies using the [Java Platform Module System (JPMS)](https://en.wikipedia.org/wiki/Java_Platform_Module_System) which was introduced in Java 9. There are no backward incompatible changes to the syntax, so in terms of the "the real code", i.e. our `*.java` files, I don't expect many changes.

AntoniusW:
TBH, I think it is an advantage if our software is able to run on "ancient" versions of Java. I don't mind switching CI to include all available and modern versions of Java, but requiring it without some good reasons feels like unnecessary self-constraining. The problem is, once we switch to a higher version, it will get harder and harder to switch back, because of course someone will try out new features. If we decide that the new features are really worth it, then I am also in favor of switching to more modern version, but so far, I did not see those features we really want to have.

You name JPMS, which is not available in Java 8 and which -- for some reasons, I guess -- is not widely adopted. I am simply not convinced so far, that JPMS really is something we want to do and that it is worth the requirement of a higher Java version.
In other words, what are the benefits of requiring Java 17?

[...]

PR #318, opened 2021-12-14
Fix handling of aggregate literals with only a right-hand comparison term (Issue #311)

### LINKED ISSUE #311: Lower bound cannot be given after the aggregate
Putting the lower bound of an aggregate _after_ an aggregate instead of _before_ it results in a `NullPointerException`.

For example, the following program:
```
p(1).
p(2).
q :- #count { N : p(N) } > 2.
```
results in:
```
Exception in thread "main" java.lang.NullPointerException
        at at.ac.tuwien.kr.alpha.grounder.transformation.aggregates.AggregateOperatorNormalization.rewriteAggregateOperator(AggregateOperatorNormalization.java:81)
        at at.ac.tuwien.kr.alpha.grounder.transformation.aggregates.AggregateOperatorNormalization.rewriteLiteral(AggregateOperatorNormalization.java:63)
        at at.ac.tuwien.kr.alpha.grounder.transformation.aggregates.AggregateOperatorNormalization.normalize(AggregateOperatorNormalization.java:56)
        at at.ac.tuwien.kr.alpha.grounder.transformation.aggregates.AggregateRewriting.apply(AggregateRewriting.java:77)
        at at.ac.tuwien.kr.alpha.grounder.transformation.NormalizeProgramTransformation.apply(NormalizeProgramTransformation.java:31)
        at at.ac.tuwien.kr.alpha.api.Alpha.normalizeProgram(Alpha.java:125)
        at at.ac.tuwien.kr.alpha.Main.main(Main.java:98)
```

Related to #268.

madmike200590:
Fixed in `AggregateOperatorNormalization`:
Aggregates of form `#aggr{...} OP TERM` are rewritten into `TERM inv(OP) #aggr{...}` and then processed as before.
`inv(OP)` refers to the inverse operator, i.e. `inv(<) = >=` etc. For operators `=` and `!=` no inverse operator is used, i.e. rewritten form is `TERM OP #aggr{...}`.

Closes #311

### DIFF
--- a/alpha-core/src/main/java/at/ac/tuwien/kr/alpha/core/programs/transformation/aggregates/AggregateOperatorNormalization.java
+++ b/alpha-core/src/main/java/at/ac/tuwien/kr/alpha/core/programs/transformation/aggregates/AggregateOperatorNormalization.java
@@ -70,14 +70,22 @@ public final class AggregateOperatorNormalization {
 	private static List<Literal> rewriteAggregateOperator(AggregateLiteral lit) {
 		AggregateAtom atom = lit.getAtom();
 		if (atom.getLowerBoundOperator() == null && atom.getUpperBoundOperator() != null) {
-			return rewriteAggregateOperator(
-				Atoms.newAggregateAtom(
+			ComparisonOperator operator = atom.getUpperBoundOperator();
+			AggregateLiteral flipped; // "flip" right-hand operator and term to left-hand ones.
+			if (operator.equals(ComparisonOperators.EQ) || operator.equals(ComparisonOperators.NE)) {
+				flipped = Atoms.newAggregateAtom(
+					atom.getUpperBoundOperator(), 
+					atom.getUpperBoundTerm(), 
+					atom.getAggregateFunction(), 
+					atom.getAggregateElements()).toLiteral(!lit.isNegated());
+			} else {
+				flipped = Atoms.newAggregateAtom(
 					atom.getUpperBoundOperator().negate(), 
 					atom.getUpperBoundTerm(), 
 					atom.getAggregateFunction(), 
-					atom.getAggregateElements())
-					.toLiteral(!lit.isNegated())
-				);
+					atom.getAggregateElements()).toLiteral(!lit.isNegated());
+			}
+			return rewriteAggregateOperator(flipped);
 		}
 		if (lit.getAtom().getAggregateFunction() == AggregateFunctionSymbol.MIN || lit.getAtom().getAggregateFunction() == AggregateFunctionSymbol.MAX) {
 			// No operator normalization needed for #min/#max aggregates.
--- a/alpha-core/src/main/java/at/ac/tuwien/kr/alpha/core/programs/transformation/aggregates/AggregateOperatorNormalization.java
+++ b/alpha-core/src/main/java/at/ac/tuwien/kr/alpha/core/programs/transformation/aggregates/AggregateOperatorNormalization.java
@@ -39,7 +39,8 @@ import at.ac.tuwien.kr.alpha.core.rules.BasicRule;
  * </ul>
  * Operators for "#min" and "#max" aggregates are not rewritten.
  * 
- * Note that input programs must only contain aggregate literals of form <code>TERM OP #aggr{...}</code> or <code>#aggr{...} OP TERM</code>, i.e. with only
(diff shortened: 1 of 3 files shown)


alpha-core/src/test/java/at/ac/tuwien/kr/alpha/core/solver/AggregatesTest.java
```diff
@@ -335,4 +335,36 @@ public void aggregateSumGlobalVariable(RegressionTestConfig cfg) {
 				"full(2)");
 	}
 
+	@AggregateRegressionTest
+	public void aggregateRightHandTermCountGreater(RegressionTestConfig cfg) {
+		String program = "p(1)." +
```
lorenzleutgeb, line 340: @AntoniusW regarding #301: Here, IMO [JEP-378: Text Blocks](https://openjdk.java.net/jeps/378) would make this look prettier.

alpha-core/src/main/java/at/ac/tuwien/kr/alpha/core/programs/transformation/aggregates/AggregateOperatorNormalization.java
```diff
@@ -69,6 +69,24 @@ private AggregateOperatorNormalization() {
...
+					atom.getUpperBoundTerm(), 
+					atom.getAggregateFunction(), 
+					atom.getAggregateElements()).toLiteral(!lit.isNegated());
+			} else {
+				flipped = Atoms.newAggregateAtom(
+					atom.getUpperBoundOperator().negate(), 
+					atom.getUpperBoundTerm(), 
+					atom.getAggregateFunction(), 
+					atom.getAggregateElements()).toLiteral(!lit.isNegated());
+			}
```
lorenzleutgeb, line 87: The difference in the second argument seems to be the only difference between the two branches. You could make that more explicit by doing:

```suggestion
			var flippedUpperBoundOperator = EQ.equals(operator) || NE.equals(operator) ? atom.getUpperBoundOperator() : atom.getUpperBoundOperator().negate();
                        // "flip" right-hand operator and term to left-hand ones.
			AggregateLiteral flipped = Atoms.newAggregateAtom(
					flippedUpperBoundOperator, 
					atom.getUpperBoundTerm(), 
					atom.getAggregateFunction(),
					atom.getAggregateElements()).toLiteral(!lit.isNegated());
```

lorenzleutgeb (approved):
LGTM, minor remarks.

AntoniusW:
> Fixed in `AggregateOperatorNormalization`: Aggregates of form `#aggr{...} OP TERM` are rewritten into `TERM inv(OP) #aggr{...}` and then processed as before. `inv(OP)` refers to the inverse operator, i.e. `inv(<) = >=` etc. For operators `=` and `!=` no inverse operator is used, i.e. rewritten form is `TERM OP #aggr{...}`.

Uhm, I beg to disagree:
`#agg {..} > 2` should be equal to `2 < #agg {..}`, i.e., what is needed here is flipping of operators, not negation.

Consider for example `X < Y` which is flipped `Y > X` and not `Y >= X`, to see that consider the case with `X=2,Y=2` and we have `2 < 2` is false as is `2 > 2` while `2 >= 2` becomes true. The unit tests currently do not consider such a case and hence cannot show that the implementation is bugged at the moment.

Please change that.

rtaupe (changes requested):
Please address @AntoniusW's comment.

madmike200590:
> > Fixed in `AggregateOperatorNormalization`: Aggregates of form `#aggr{...} OP TERM` are rewritten into `TERM inv(OP) #aggr{...}` and then processed as before. `inv(OP)` refers to the inverse operator, i.e. `inv(<) = >=` etc. For operators `=` and `!=` no inverse operator is used, i.e. rewritten form is `TERM OP #aggr{...}`.
> 
> Uhm, I beg to disagree: `#agg {..} > 2` should be equal to `2 < #agg {..}`, i.e., what is needed here is flipping of operators, not negation.
> 
> Consider for example `X < Y` which is flipped `Y > X` and not `Y >= X`, to see that consider the case with `X=2,Y=2` and we have `2 < 2` is false as is `2 > 2` while `2 >= 2` becomes true. The unit tests currently do not consider such a case and hence cannot show that the implementation is bugged at the moment.
> 
> Please change that.

Thanks for catching this! Since the actual rewriting in `AggregateOperatorNormalization` uses operator negation in quite a few places, I mentally tripped myself up here :sweat_smile: 

This should now be fixed.

rtaupe (approved):
Thanks for fixing this! 😃

PR #325, opened 2021-12-17
release preparation: set version to 0.6.0

madmike200590:
In preparation for the actual release of version 0.6.0, we need to set the correct version in gradle.properties before we can create the tag


madmike200590:
@lorenzleutgeb unfortunately I cannot do this in `master` directly due to the workflow rule requiring reviewed PRs.

AntoniusW (approved):

lorenzleutgeb (approved):

PR #331, opened 2021-12-20
Build Properties

lorenzleutgeb:
Introduce a mechanism to pass "Build Properties" (version number, hash of the source revision, etc.) to the resulting artifact. Mainly to allow Alpha report its own version, which is implemented as example.

This contents of `META-INF/build.properties` must be kept reproducible, i.e. only depend on our Git repository as input, not use timestamps, current hostname or weather...

### DIFF
--- a/alpha-cli-app/build.gradle.kts
+++ b/alpha-cli-app/build.gradle.kts
@@ -40,6 +40,10 @@ tasks.create<Jar>("bundledJar") {
 	archiveFileName.set("${project.name}-${project.version}-bundled.jar")
 
 	exclude("META-INF/DEPENDENCIES")
+
+	filesMatching("META-INF/build.properties") {
+		duplicatesStrategy = DuplicatesStrategy.INCLUDE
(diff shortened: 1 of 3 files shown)


madmike200590 (commented):
Works as expected so far, `-version` CLI flag correctly outputs `build.properties` content.

However, we have duplicate zip entries for the `build.properties` file (gets created once by each module). This has two immediate consequences:
- Our jar files become invalid for `unzip`, i.e. unzipping of the bundled jar is no longer possible (see [this](https://github.com/gradle/gradle/issues/9990) issue for some background information) due to these duplicate entries.
- In a setting where different modules of alpha get packaged into one jar that do not share the exact same version, an overwrite-strategy for duplicate handling would destroy information (likelyhood of that seems pretty low).

While I'm not too concerned about the second item, I think we do need a fix for the invalid zip issue.

lorenzleutgeb:
Right. Thanks for catching this.

PR #336, opened 2022-02-08
Refactoring: Decouple API implementation from config structures

madmike200590:
The aim of this PR is to get rid of all dependencies between `SystemConfig` and API implementations (like `AlphaImpl`, `NaiveGrounder`, `DefaultSolver` etc).
The idea is that these objects should get their dependencies at the time of construction from an external source (`AlphaFactory` as it is implemented now) and not need to be aware of any configuration mechanisms or structures themselves.

The same mechanism of obtaining `Alpha` instances (only through `AlphaFactory`) has been applied to all tests that test complete workflows (i.e. tests constituting end-to-end-tests in that they test the complete system rather than a single unit of code) as well. These tests (everything annotated `RegressionTest` or `AggregateRegressionTest`) have been moved to the `alpha-solver` module and adapted to be parameterized using a `SystemConfig` which is used to obtain an `Alpha` instance from the factory.

Additional changes:
- renamed `ASPCore2Program` to `InputProgram`
- removed config option for grounder name; it doesn't seem realistic that we get an alternative implementation in the short- or mid-term that would make such an option necessary
- removed "deterministic" CLI option - instead select a seed directly
- pulled code used in tests across multiple modules into separate source folder `alpha-core/src/testFixtures` that can be shared between modules
- fix minor bug in `StratifiedEvaluation`: Avoid generating duplicate facts by using a `Set` rather than `List` as intermediate storage for newly derived rule heads


alpha-solver/src/test/java/at/ac/tuwien/kr/alpha/regressiontests/SolverStatisticsTests.java
```diff
@@ -51,32 +54,33 @@ public void setUp() {
...
 		collectAnswerSetsAndCheckStats(solver, 2, 1, 1, 1, 1, 0, 0, 0);
 	}
 
-	@RegressionTest
-	public void checkNoGoodCounterStatsByTypeUsingDummyGrounder(RegressionTestConfig cfg) {
-		Solver solver = buildSolverForRegressionTest(atomStore, new GrounderMockWithBasicProgram(atomStore), cfg);
-		assumeTrue(solver instanceof StatisticsReportingSolver);
-		collectAnswerSetsAndCheckNoGoodCounterStatsByType(solver, 4, 0, 0, 0);
-	}
+	// TODO Why are these tests config-dependent if they use a grounder mock and seem rather like unit tests on solver statistics?
```
madmike200590, line 70: @AntoniusW see `TODO`: I don't see a reason for these tests to be part of this class, i.e. I think these should be unit tests (and not parameterized with different configs). Let me know if you agree and I can move those to an appropriate class in the `alpha-core` module.

alpha-solver/src/test/java/at/ac/tuwien/kr/alpha/regressiontests/SolverTests.java
```diff
@@ -805,16 +803,17 @@ public void smallCardinalityAggregate(RegressionTestConfig cfg) {
 		);
 	}
 
-	@RegressionTest
-	public void dummyGrounder(RegressionTestConfig cfg) {
-		AtomStore atomStore = new AtomStoreImpl();
-		assertEquals(GrounderMockWithBasicProgram.EXPECTED, buildSolverForRegressionTest(atomStore, new GrounderMockWithBasicProgram(atomStore), cfg).collectSet());
-	}
+	// TODO @AntoniusW what are these? Can we get rid of them? If not, where do I move them?
```
madmike200590, line 806: @AntoniusW see `TODO`: The commented-out test cases don't seem to be following the pattern of an end-to-end system test of all other tests in this class. Should we move these to some other location and remove the parameterization?

Issue #346, opened 2022-09-11
Change SimpleLogger time format


paulbehofsics:
In the comments of the `simplelogger.properties` files the number of milliseconds since startup is specified as the intended time format for logging. The current configuration results in the number of milliseconds since the last second.

```
# Format for logged timestamps. 
# For Alpha, we use milliseconds since startup
org.slf4j.simpleLogger.dateTimeFormat=S
```

PR #347, opened 2022-09-11
Fix SimpleLogger time format

paulbehofsics:
Related issue: #346


madmike200590 (approved):
Thanks for catching (and fixing) this!

PR #353, opened 2023-01-03
Get rid of String-encoding of Substitutions in RuleAtoms

### LINKED ISSUE #352: Cannot reconstruct substitution with function terms correctly from body-representing atoms
Body-representing atoms, i.e., `RuleAtom`-s, currently use strings to represent the substitution with which the non-ground rule was grounded with. The `String`-representation has always been ugly and in addition leads to a bug when the original `Substitution` is reconstructed and if the substitution contained function terms. The following test causes the bug on the master branch.

```
	@Test
	public void substitutionWithFunctionTermsObtainableFromRuleAtom() {
		Rule<Head> rule = PARSER.parse("q(X) :- p(X,Y).").getRules().get(0);
		CompiledRule nonGroundRule = InternalRule.fromNormalRule(NormalRuleImpl.fromBasicRule(rule));
		// Build substitution X -> b(a,a), Y -> b(b(a,a),b(a,a)).
		BasicAtom atomForSpecialize = Atoms.newBasicAtom(Predicates.getPredicate("p", 2), X, Y);
		ConstantTerm<String> aTerm = Terms.newSymbolicConstant("a");
		Instance instanceForSpecialize = new Instance(Terms.newFunctionTerm("b", aTerm, aTerm),
			Terms.newFunctionTerm("b",
				Terms.newFunctionTerm("b", aTerm, aTerm),
				Terms.newFunctionTerm("b", aTerm, aTerm)));
		Substitution substitution = BasicSubstitution.specializeSubstitution(
			atomForSpecialize, instanceForSpecialize, BasicSubstitution.EMPTY_SUBSTITUTION);

		RuleAtom ruleAtom = new RuleAtom(nonGroundRule, substitution);
		String substitutionString = (String) ((ConstantTerm<?>) ruleAtom.getTerms().get(1)).getObject();
		Substitution fromString = Substitutions.fromString(substitutionString);
		assertEquals(substitution, fromString);
	}
```
Although the buggy code is in `Substitutions.fromString(..)` and could be fixed using a better parser, an even better solution seems to be getting rid of the `String`-representation itself and using instead of the `String`, a `ConstantTerm<..>` that directly refers to the `Substitution` object.

AntoniusW:
Removes the encoding of `Substitution`s as a `String` inside body-representing atoms, i.e., `RuleAtom`s. Switches over to using `ConstantTerm<RuleAtomData>` where the `Substitution` is directly contained in the `RuleAtomData`. Fixes #352.


alpha-core/src/main/java/at/ac/tuwien/kr/alpha/core/atoms/RuleAtom.java
```diff
@@ -43,27 +38,92 @@
...
-		this.terms = terms;
+		@Override
+		public int compareTo(RuleAtomData other) {
+			if (nonGroundRule.getRuleId() != other.nonGroundRule.getRuleId()) {
+				return Integer.compare(nonGroundRule.getRuleId(), other.nonGroundRule.getRuleId());
+			}
+			if (substitution.getSubstitution().size() != other.getSubstitution().getSubstitution().size()) {
+				throw oops("RuleAtoms over the same rule have different-sized substitutions.");
+			}
+			// Note: We assume here that substitutions for the same nonGroundRule are all over the same variables and iteration order.
```
madmike200590, line 76: In addition to having to rely on this assumption, there is also never a point where a connection is made from the variables set in the `Substitution`s vs. the ones actually occurring in the rule. See PR #354 for a suggestion of a more robust version that also doesn't make this assumption.

madmike200590 (commented):
Looks nice! I've got one suggestion, see PR #354, otherwise ok with me.

alpha-core/src/main/java/at/ac/tuwien/kr/alpha/core/atoms/RuleAtom.java
AntoniusW: Thank you for pointing at this issue and suggesting a solution to it. After some deliberation, I changed the code in another way for the following reasons:
- The suggested code repeatedly collects all occurring variables of the underlying non-ground rule (every time `compareTo` is called).
- If we really want to ensure that a given `Substitution` fits the `CompiledRule`, it should be done in the constructor and not on a call to `compareTo`.
- The `RuleAtom` is in the core package, so the condition that substitutions fit the non-ground rule should never be violated.
- The assumption mentioned in the comment above, about variable iteration order, is actually always true as the `Substitution` interface guarantees that the substitution is backed by a TreeMap, which has a well-defined iteration order.
- The other part of the assumption (that substitutions are over the same variables) can easily be checked. The changed code does that now.

Please have a look at the updated code and feel free to criticise my reasoning above.

madmike200590 (approved):
Looks good to me!

paulbehofsics (approved):
Looks good!

PR #354, opened 2023-01-03
Revise RuleAtomData#compareTo

madmike200590:
Eliminates the implicit assumption for two compared `RuleAtomData`s to have the same variables set and instead works off the occurring variables of the rule.


madmike200590:
Proposed change made obsolete by better solution.

PR #355, opened 2023-01-10
ci: Upgrade actions/setup-java

lorenzleutgeb:
To work around deprecation of `set-output`, see also https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


madmike200590 (approved):
Looks good, thanks for catching this!

PR #356, opened 2023-01-21
Add Jacoco aggregation for coverage.

AntoniusW:
Refactoring Alpha into modules resulted in coverage to be computed only individually for each module. This results in many lines of code being considered not hit by any test, while they indeed are tested in (functional-/integration-)tests of another module. This PR tries to rectify coverage computation by using Jacoco's aggregate plugin on the top-level of the build.


AntoniusW:
@lorenzleutgeb Since you are the one with the most Gradle experience in our team, could you have a look at this PR and check it? Thanks!

lorenzleutgeb (approved):

PR #359, opened 2023-03-07
Make releasing Alpha more convenient by adding gradle release plugin

madmike200590:
Adds a basic configuration for the [Gradle Release Plugin](https://github.com/researchgate/gradle-release) to `build.gradle.kts`. The release plugin
- Sets version in `gradle.properties` to the value passed to the plugin
- Creates a tag with the release version
- Sets version in `gradle.properties` of `master` branch to the next version passed to the plugin

#### Example
Running `./gradlew release -Prelease.useAutomaticVersion=true -Prelease.releaseVersion=0.7.0 -Prelease.newVersion=0.8.0-SNAPSHOT`  would create a tag `0.7.0` and set the version in master to `0.8.0-SNAPSHOT`.

The plugin will fail if it is run on a branch other than `master` or if there are unversioned files present in the working directory.

#### Permissions
The release plugin requires permissions to directly push to the `master` branch. It can therefore only be run by users that are allowed to bypass the branch protection rule for `master` requiring approved pull requests to push. By default, these are repository admins and owners, including owners of the organization owning the repository (who are transitively repo owners).


automaticRelease.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
AntoniusW, line 1: Where is this script called from?

automaticRelease.sh
```diff
@@ -0,0 +1,50 @@
...
+function prop {
+    grep "${1}" ${file} | cut -d'=' -f2
+}
+
+version=$(prop 'version')
+release_major=0
+release_minor=0
+release_patch=0
+
+# break down the version number into it's components
```
AntoniusW, line 15:
```suggestion
# Break down the version number into its components
```

automaticRelease.sh
```diff
@@ -0,0 +1,50 @@
...
+regex="([0-9]+).([0-9]+).([0-9]+)-SNAPSHOT"
```
AntoniusW, line 16: This always expects versions to end with "-SNAPSHOT", is that the intended behaviour?

.github/workflows/release.yml
```diff
@@ -0,0 +1,42 @@
...
+    - name: Prepare git config
+      run: |
+        git config --local user.name alpha-release-action
+        git config --local user.email alpha-release-action@no.where    
+    - name: Run Gradle Release Plugin
+    # See https://github.com/marketplace/actions/gradle-build-action
+    # https://github.com/researchgate/gradle-release  
+      uses: gradle/gradle-build-action@v2
+      with:
+        arguments: release -Prelease.useAutomaticVersion=true -Prelease.releaseVersion=${{ github.event.inputs.release_version}} -Prelease.newVersion=${{ github.event.inputs.next_version}}
```
AntoniusW, line 41: If there is an easy way to make sure that release versions are increasing only, I'd like to see that used. But I guess that is really hard to achieve in cases where, e.g., a new patch is back-ported to an older version. Is there some way to exclude that version numbers are duplicated?

automaticRelease.sh
madmike200590: Actually nowhere, that's a leftover from an unsuccessful attempt to automatically generate version identifiers. I deleted the script.

automaticRelease.sh
madmike200590: See above, that script was a leftover

.github/workflows/release.yml
madmike200590: See above comments about my failed attempt to do that via script.
Log story short, I think it's more trouble than it's worth: Since you may have perfectly valid reasons to release pretty much any version at any point in time, I don't think such a check could ever be "smart enough" to be useful. Even if your current development version (i.e. version in `master`) were `2.3.0`, you could still find a bug in version `1.7.0` that needs urgent fixing but is not easily ported to your current version, thus requiring you to release a version `1.7.1` that branches from your old tag.
Virtually the only useful check on release versions is if you already have a tag for the same version - and that happens automatically since git won't let you create duplicate tags.

madmike200590:
> Regarding the general setup here: is it possible to not push to `master` but have a separate `releases` branch where the release bot then has write access to?
That's possible, but I think it would defeat the purpose of this PR (see below).

> At the moment, our `master` is a hand-checked version of the code, hence I have great trust in it. As soon as we start to see automatic commits, this trust is gone -- obviously, we can still check all the commits done by the script after the fact, but at some point no one will do that any longer. 
I don't see the trust issue. We use a fixed version of an established plugin (gradle release plugin in version 3.0.2) that we call through a validated instance of the gradle wrapper (ensured through wrapper validation action in the workflow). All the code executed during an automated release is publicly accessible and we can go through it with a fine-toothed comb anytime if we ever feel the need to. So, in my opinion the automated release using this workflow is about the safest possible way to do it and way more reliable than a human manually executing these steps. 

> So, I would prefer the script for releases not to commit any source to master.
For the sake of argument, let's lay out the workflow we'd get if we release through a separate branch. I'm assuming both `master` and `releases` to be protected, i.e. both can only be pushed to using a pull request. In addition, `releases` can be pushed to by some `bot` user that's only used in the release action.
1.) Person A wants to release a new version. They create a PR from `master` to `releases`.
2.) Someone else, Person B, needs to approve the release PR.
3.) Either A or B triggers the release action, a tag from `releases` is created, the actual release happens.
4.) A now needs to open a new PR on `master`, merging back the state from `releases` in order to have the new `SNAPSHOT` version in `master`.
All the while, no new PR must be merged into master because that would change the scope of the release (since the "release" tagged from `releases` does not contain what A wanted to release, but whatever was on master when B merged the PR).
A possible way around that would be to have `releases` unprotected - but then I'd argue that just simply having the release action work in `master` is the safer way of going about this. And even that would not eliminate the need for the PR in step 4.

Since the idea behind this PR was to get to a less cumbersome release process that does not involve pull requests and two people, I'd very much like to do it the way I suggested.

madmike200590:
Changed the PR to only include config for gradle release plugin since I didn't find any way to have a release workflow in github actions without creating a personal access token that would be tied to my own account (or one managed by me) with way more permissions than I feel comfortable with.

madmike200590:
@AntoniusW as discussed offline, I am merging this as-is in preparation for the next release of Alpha.

